PR #5, opened 2018-08-17
Handle JobId

samuel-dev-dba:
Pass jobId to the outgoing message.


src/SFA.DAS.Payments.RequiredPayments.RequiredPaymentsService/RequiredPaymentsService.cs
```diff
@@ -59,7 +60,11 @@ public async Task<CalculatedPaymentDueEvent[]> HandlePayableEarning(PayableEarni
 
                 var paymentsDue = apprenticeship.CreatePaymentDue(new[] { earning }, paymentHistory);
 
-                return paymentsDue.Select(_mapper.Map<PaymentDue, CalculatedPaymentDueEvent>).ToArray();
+                var calculatedPaymentDueEvent = paymentsDue.Select(_mapper.Map<PaymentDue, CalculatedPaymentDueEvent>).ToList();
+
+                calculatedPaymentDueEvent.ForEach(x => x.JobId = earningEntity.JobId);
```
abroodie: Can setting the job id be done in the mapper?

abroodie: This code will be moved into a handler and we will also handle the JobId a little differently.

PR #29, opened 2018-09-20
Pv2 56 payment funding source handle act2 on programme

samuel-dev-dba:


kuzinator:
if you've ignored App.Release.config, shouldn't it be taken off source control?

PR #34, opened 2018-09-27
PV2-219 - addition of missing specflow steps and tidying up of steps …

davehowlett:
…that were not consistently named

### DIFF
--- a/src/SFA.DAS.Payments.RequiredPayments.AcceptanceTests/Steps/RequiredPaymentsSteps.cs
+++ b/src/SFA.DAS.Payments.RequiredPayments.AcceptanceTests/Steps/RequiredPaymentsSteps.cs
@@ -48,14 +48,14 @@ namespace SFA.DAS.Payments.RequiredPayments.AcceptanceTests.Steps
             return payment;
         }
 
-        [Then(@"the required payments component will generate the following contract type (.*) payable earnings:")]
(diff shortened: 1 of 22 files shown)


src/SFA.DAS.Payments.RequiredPayments.AcceptanceTests/Steps/RequiredPaymentsSteps.cs
```diff
@@ -48,40 +48,78 @@ private ApprenticeshipContractTypePaymentDueEvent CreatePaymentDueEvent(Data.OnP
...
                         && expectedEvent.Type == receivedEvent.OnProgrammeEarningType
                         && TestSession.Ukprn == receivedEvent.Ukprn
                         && expectedEvent.Period == receivedEvent.DeliveryPeriod?.Period
                         && receivedEvent.CollectionPeriod.Name.Contains(CollectionYear)
-                        ));
+                    ));
+            }, "Failed to find all the required payment events");
+        }
+
+        [Then(@"the required payments component will generate the following contract type (.*) Learning \(TT(.*)\) payable earnings:")]
```
kuzinator: Learning is always TT1, is it? Would that be more obvious if it was "Learning (TT1)" instead of "Learning \(TT(.*)\)"?
And the same with Balancing and Completion. Alternatively to parametrise the whole type.

PR #56, opened 2018-11-12
PV2-255: non levy standard learner price is changed and a negative amount is left to be paid

kuzinator:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/RefundsSteps.cs
```diff
@@ -49,13 +49,27 @@ public void GivenTheFollowingProviderPaymentsHadBeenGenerated(Table table)
         [Given(@"the Provider now changes the Learner details as follows")]
         public void GivenTheProviderNowChangesTheLearnerDetailsAsFollows(Table table)
         {
-            var ilr = table.CreateSet<Training>().ToList();
+            var ilr = table.CreateSet<Training>().Select(t =>
+            {
+                if ((byte) t.ContractType == 0)
```
samuel-dev-dba: We should be using all input from Specflow scenario table to create a learner

src/SFA.DAS.Payments.EarningEvents.Domain.UnitTests/Validation/Learner/OverlappingPriceEpisodeValidationRuleTests.cs
```diff
@@ -42,5 +43,40 @@ public void Validation_Fails_If_There_Are_Overlapping_Price_Episodes()
 
             Assert.IsTrue(result.Failed,result.FailureReason);
         }
+
+        [Test]
+        public void Validation_Passes_If_There_Are_Multiple_Price_Episodes()
```
samuel-dev-dba: Please remove the underscore in the Test Method Name

PR #65, opened 2018-11-20
Update the project definitions to build for 'Any CPU' rather than x64.

DavidBSteele:
Added a try/catch around the ILRSubmissionService to ensure all errors are logged.


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/SFA.DAS.Payments.AcceptanceTests.EndToEnd.csproj
```diff
@@ -25,7 +25,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <PlatformTarget>x64</PlatformTarget>
+    <PlatformTarget>AnyCPU</PlatformTarget>
```
kuzinator: We found that Any CPU build will prevent this test from running on VSTS due to reference to service fabric x64 packages.

PR #66, opened 2018-11-20
Pv2 334 6 integration branch

davehowlett:


davehowlett:
to be merged from original requests

PR #68, opened 2018-11-21
Pv2 400 allow days in date format

carlbm:


src/SFA.DAS.Payments.AcceptanceTests.Core/Extensions.cs
```diff
@@ -14,6 +14,25 @@ public static CalendarPeriod ToCalendarPeriod(this DateTime date)
...
             
         public static DateTime ToDate(this string dateText)
+        {
+            var parts = dateText.Split('/');
+            if (parts.Length == 3)
+            {
+                var monthAndYear = ParseTwoPartDescription($"{parts[1]}/{parts[2]}");
+                try
+                {
+                    var result = monthAndYear.AddDays(int.Parse(parts[0]));
```
davehowlett: don't need to create a variable here. Just return it.

PR #70, opened 2018-11-21
Pv2 282 start date moves forward

carlbm:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Tests/Non-Levy Learner - Learner Change of Circs/Change start date forward within calendar month.feature
```diff
@@ -0,0 +1,90 @@
+﻿Feature: Learner Change of Circs
```
davehowlett: We include the story number on the filename and the feature heading

davehowlett: and also, the description from the ticket as well for future reference:

As a provider,

I want a non-levy learner, where change to start date within calendar month, forward in month is paid the correct amount

So that I am accurately paid my apprenticeship provision.

PR #78, opened 2018-11-26
Pv2 265 implement mapping logic of fm36 model output to earning event messages

carlbm:


src/SFA.DAS.Payments.EarningEvents.Application/Mapping/ApprenticeshipContractTypeEarningsEventBuilder.cs
```diff
@@ -17,14 +20,88 @@ public ApprenticeshipContractTypeEarningsEventBuilder(IApprenticeshipContractTyp
...
+                    .Where(x => x.PriceEpisodeValues.PriceEpisodeAimSeqNumber == learningDelivery.AimSeqNumber);
+                var intermediateAim = new IntermediateLearningAim(learnerSubmission, priceEpisodes, learningDelivery);
+                results.Add(intermediateAim);
+            }
+
+            return results;
+        }
+    }
+
+    public class IntermediateLearningAim
```
kuzinator: Would it look better in a separate file?

PR #79, opened 2018-11-27
Required Payments Cache Cleaner changed to use NSB

kuzinator:


src/SFA.DAS.Payments.RequiredPayments.RequiredPaymentsProxyService/Handlers/CollectionStartedEventHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public async Task Handle(CollectionStartedEvent message, IMessageHandlerContext context)
+        {
+            paymentLogger.LogInfo($"Processing CollectionStartedEvent event. Message Id : {context.MessageId}");
+
+            foreach (var key in message.ApprenticeshipKeys)
+            {
+                var actorId = new ActorId(key);
+                var actor = proxyFactory.CreateActorProxy<IRequiredPaymentsService>(new Uri("fabric:/SFA.DAS.Payments.RequiredPayments.ServiceFabric/RequiredPaymentsServiceActorService"), actorId);
+
```
carlbm: Being very picky here... but is it worth putting this URI into a static field accessible from the entire component? I've seen it in a few places now

src/SFA.DAS.Payments.RequiredPayments.RequiredPaymentsProxyService/Handlers/CollectionStartedEventHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+        public CollectionStartedEventHandler(IActorProxyFactory proxyFactory, IPaymentLogger paymentLogger)
+        {
+            this.proxyFactory = proxyFactory;
+            this.paymentLogger = paymentLogger;
+        }
+
+        public async Task Handle(CollectionStartedEvent message, IMessageHandlerContext context)
+        {
+            paymentLogger.LogInfo($"Processing CollectionStartedEvent event. Message Id : {context.MessageId}");
+
```
carlbm: I would prefer this sort of message moved to a pipeline stage - do it once per component and never worry again

PR #86, opened 2018-12-03
Pv2 411 payments due act2 incentives

davehowlett:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Infrastructure/FeatureScopedBootstrapper.cs
```diff
@@ -30,7 +32,8 @@ public static void SetUpContainer()
             DcHelper.AddDcConfig(Builder);
 
             Builder.RegisterType<ApprenticeshipKeyService>().AsImplementedInterfaces();
-            Builder.RegisterType<RequiredPaymentsCacheCleaner>().AsSelf();
+
+            Builder.Register((c, p) => new RequiredPaymentsCacheCleaner(c.Resolve<IApprenticeshipKeyService>(), MessageSession)).AsSelf();
```
abroodie: Why can't this use RegisterType<>?

src/SFA.DAS.Payments.PaymentsDue.Domain.UnitTests/Services/ApprenticeshipContractType2EarningProcessorTest.cs
```diff
@@ -12,20 +14,22 @@ namespace SFA.DAS.Payments.PaymentsDue.Domain.UnitTests.Services
     [TestFixture]
     public class ApprenticeshipContractType2EarningProcessorTest
```
kuzinator: This test class' name suggests that it is for ACT2 earning processor, Incentives are not ACT2 even though very similar in the way they get processed. This creates confusion further down the class where parent classes are used where ACT2 classes could be used. Is this a case where we need to create a new class for incentives and factor out common parts? There will be more processors coming.

src/SFA.DAS.Payments.PaymentsDue.Domain/ApprenticeshipContractTypeIncentiveProcessor.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using SFA.DAS.Payments.Model.Core;
+using SFA.DAS.Payments.Model.Core.Incentives;
+using SFA.DAS.Payments.PaymentsDue.Messages.Events;
+
+namespace SFA.DAS.Payments.PaymentsDue.Domain
+{
+    public class ApprenticeshipContractTypeIncentiveProcessor : IApprenticeshipContractTypeIncentiveProcessor
```
abroodie: I think this class can be used to build incentives in response to FunctionalSkills events (Math & English) too.  Maybe it should be renamed to IncentiveProcessor instead?

src/SFA.DAS.Payments.PaymentsDue.Domain/ApprenticeshipContractTypeIncentiveProcessor.cs
```diff
@@ -0,0 +1,54 @@
...
+            var paymentsDue = new List<IncentivePaymentDueEvent>();
+
+            foreach (var period in incentiveEarning.Periods.Where(earning => earning.Period <= collectionPeriod.Period))
+            {
+                paymentsDue.Add(new IncentivePaymentDueEvent
+                {
+                    JobId = jobId,
+                    Ukprn = ukprn,
+                    DeliveryPeriod =
+                        new CalendarPeriod(collectionPeriod.Name.Split('-').FirstOrDefault(),
```
kuzinator: Bafore we decide if or how to refactor CalendarPeriod class, it could benefit from "AcademicYear" property that will do exactly what is done here.

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Infrastructure/FeatureScopedBootstrapper.cs
kuzinator: Because it uses MessageSession that is initialised further down the order of steps. This is the simplest way to achieve it without passing message session to every call or refactoring steps base.

src/SFA.DAS.Payments.PaymentsDue.Domain/ApprenticeshipContractTypeIncentiveProcessor.cs
davehowlett: pushed to a subtask of PV2-395

src/SFA.DAS.Payments.PaymentsDue.Domain.UnitTests/Services/ApprenticeshipContractType2EarningProcessorTest.cs
davehowlett: changed - refactored existing class and added an IncentiveProcessorTest class

src/SFA.DAS.Payments.PaymentsDue.Domain/ApprenticeshipContractTypeIncentiveProcessor.cs
davehowlett: Renamed

PR #95, opened 2018-12-10
Pv2 394 learner changes uln

carlbm:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndSteps.cs
```diff
@@ -96,13 +96,21 @@ public void GivenPriceDetailsAsFollows(Table table)
         public async Task ThenTheFollowingLearnerEarningsShouldBeGenerated(Table table)
         {
             var earnings = table.CreateSet<OnProgrammeEarning>().ToList();
+            foreach (var earning in earnings)
```
samuel-dev-dba: Why are we doing this as I believe we will be using what has been set in the Test Session to populate the leaner. So no need to  set it directly

ESFA-Pawan:
Retested in DAS Payments V2-0.0.0701-R002

ESFA-Pawan:
@kuzinator this has been already merged to master. If any further changes are needed then they should be done as separate story.

PR #98, opened 2018-12-11
PV2-458 Added step to change to a new provider

carlbm:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndSteps.cs
```diff
@@ -42,6 +42,21 @@ public void GivenTheProviderPreviouslySubmittedTheFollowingLearnerDetailsInColle
...
+            var ilr = table.CreateSet<Training>().ToList();
+            CurrentIlr = ilr;
+            AddTestLearners(CurrentIlr);
+        }
+
+        [Given("the Learner has now changed to \"(.*)\" as follows")]
+        public void GivenTheLearnerChangesProvider(string providerId, Table table)
+        {
+            TestSession.RegenerateUkprn();
+            GivenTheProviderNowChangesTheLearnerDetailsAsFollows(table);
```
samuel-dev-dba: I will suggest we move common test code to the base class and share it rather than calling test step methods within another test method.

carlbm:
Updated base branch to the feature branch

PR #101, opened 2018-12-12
Pv2 324 change provider with a gap

carlbm:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/RefundsSteps.cs
```diff
@@ -19,6 +20,14 @@ public RefundsSteps(FeatureContext context) : base(context)
         {
         }
 
+        [Given("\"(.*)\" previously submitted the following learner details")]
+        public void GivenTheProviderPreviouslySubmittedTheFollowingLearnerDetailsForProvider(
+            string providerId,
+            Table table)
+        {
+            GivenTheProviderPreviouslySubmittedTheFollowingLearnerDetails(table);
```
samuel-dev-dba: Can we push shared methods to base class rather than calling Test Method in another test method?

PR #104, opened 2018-12-20
Pv2 476 accetance test handles all incentive type

samuel-dev-dba:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Data/Payment.cs
```diff
@@ -7,6 +11,7 @@ public class Payment
         public decimal OnProgramme { get; set; }
         public decimal Completion { get; set; }
         public decimal Balancing { get; set; }
-        public decimal LearningSupport { get; set; }
+
+        public IDictionary<IncentiveType, decimal> IncentiveValues { get; set; } = new Dictionary<IncentiveType, decimal>();
```
carlbm: Rename this to 'Values' so that when we put maths and english or the on-prog in the future there's less to do?

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/EventMatchers/RequiredPaymentEventMatcher.cs
```diff
@@ -38,44 +38,58 @@ protected override IList<RequiredPaymentEvent> GetExpectedEvents()
...
+                        AmountDue = payment.Completion,
+                        OnProgrammeEarningType = OnProgrammeEarningType.Completion,
+                        DeliveryPeriod = payment.DeliveryPeriod.ToCalendarPeriod()
+                    };
                     expectedPayments.Add(completionPayment);
+                }
+
+                foreach (var incentiveTypeKey in payment.IncentiveValues.Keys)
+                {
+                    if (!payment.IncentiveValues.TryGetValue(incentiveTypeKey, out var amount)) continue;
```
carlbm: You know that the key exists and this is not a concurrent collection. There's no reason not to use payment.IncentiveValues[incentiveTypeKey]

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Data/Payment.cs
samuel-dev-dba: The plan is to have another property called MathsAndEngValues not to combine them into one.
This has an advantage when processing then in the matchers

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/EventMatchers/RequiredPaymentEventMatcher.cs
samuel-dev-dba: Changed

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Data/Payment.cs
abroodie: This is fine as Maths&English are treated as incentives by PaymentsDue, RequiredPayments, FundingSource & ProviderPayments.

PR #134, opened 2019-01-16
PV2-556 ACT1 Earning Events

kuzinator:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndStepsBase.cs
```diff
@@ -388,8 +388,7 @@ int CalculateAimSequenceNumber(Price priceEpisode)
 
             string CalculateContractType(Price priceEpisode)
             {
```
carlbm: Not sure that this is correct...

carlbm: This is not correct,  can you undo this change and replicate it through the other files changed

kuzinator: done

PR #145, opened 2019-01-24
Pvc 546 acceptance tests handle act1 earning events

samuel-dev-dba:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/EventMatchers/BaseEarningEventMatcher.cs
```diff
@@ -255,5 +220,80 @@ private static bool MatchEarningPeriods(ReadOnlyCollection<EarningPeriod> actual
...
+                {
+                    Amount = e.Values[tt],
+                    Period = e.DeliveryCalendarPeriod,
+                    PriceEpisodeIdentifier = FindPriceEpisodeIdentifier(e.Values[tt], e, fm36Learner, tt)
+                }).ToList().AsReadOnly()
+            }).ToList().AsReadOnly();
+            return earningsEvent;
+        }
+
+        protected ApprenticeshipContractTypeEarningsEvent MapToApprenticeshipContractTypeEarningsEventWithIncentives(
```
kuzinator: This is always called after an object is created. No need to make it protected and called from derived classes.

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/EventMatchers/BaseEarningEventMatcher.cs
```diff
@@ -255,5 +220,80 @@ private static bool MatchEarningPeriods(ReadOnlyCollection<EarningPeriod> actual
 
             return true;
         }
+
+        protected ApprenticeshipContractTypeEarningsEvent MapToApprenticeshipContractTypeEarningsEventWithoutIncentives(
```
kuzinator: This is always called after an object is created. No need to make it protected and called from derived classes.

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/EventMatchers/ApprenticeshipContractType1EarningEventMatcher.cs
```diff
@@ -0,0 +1,53 @@
...
+using SFA.DAS.Payments.Model.Core.Entities;
+using SFA.DAS.Payments.Model.Core.Incentives;
+using SFA.DAS.Payments.Model.Core.OnProgramme;
+using Earning = SFA.DAS.Payments.AcceptanceTests.EndToEnd.Data.Earning;
+
+namespace SFA.DAS.Payments.AcceptanceTests.EndToEnd.EventMatchers
+{
+    public class ApprenticeshipContractType1EarningEventMatcher: BaseEarningEventMatcher
+    {
+        private readonly IList<Earning> earningSpecs;
```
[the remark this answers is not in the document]
samuel-dev-dba: There are only 3 methods in the derived classes which are used in the based mather to validate events. We need to add payments to be validated so you need an object, but you cannot create an instance of the base event ApprenticeshipContractTypeEarningsEvent hence need for the derived classes. Happy to change it if there is a better way.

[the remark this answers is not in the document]
davehowlett: I agree - lets discuss. We talked about this with Tony yesterday afternoon

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/EventMatchers/EarningEventMatcher.cs
```diff
@@ -90,24 +95,23 @@ protected override IList<EarningEvent> GetExpectedEvents()
 
                     if (aimSpec.AimReference == "ZPROG001" && onProgEarnings.Any())
                     {
-                        var onProgEarning = new ApprenticeshipContractType2EarningEvent
+                        var onProgEarning  = CreateContractTypeEarningsEventEarningEvent();
+                        onProgEarning.CollectionPeriod = collectionPeriod;
```
davehowlett: the collection period and ukprn could be moved into the CreateContractTypeEarningsEventEarningEvent(); method as its common between both usages

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/EventMatchers/EarningEventMatcher.cs
```diff
@@ -90,24 +95,23 @@ protected override IList<EarningEvent> GetExpectedEvents()
...
-                            JobId = testSession.JobId,
-                            Learner = learner,
-                            LearningAim = learningAim
-                        };
+                                Amount = e.Values[tt],
+                                Period = (byte) e.DeliveryCalendarPeriod,
+                                PriceEpisodeIdentifier = FindPriceEpisodeIdentifier(e.Values[tt], e, fm36Learner, tt)
+                            }).ToList().AsReadOnly()
+                        }).ToList().AsReadOnly();
+                        onProgEarning.JobId = testSession.JobId;
```
davehowlett: These three variables could be moved to CreateContractTypeEarningsEventEarningEvent() as they are common to both usages

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/EventMatchers/EarningEventMatcher.cs
samuel-dev-dba: Done

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/EventMatchers/EarningEventMatcher.cs
samuel-dev-dba: Done

PR #152, opened 2019-01-28
Pv2 547 handle act1 required payment events

davehowlett:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/EventMatchers/RequiredPaymentEventMatcher.cs
```diff
@@ -107,5 +111,44 @@ private bool MatchIncentive(IncentiveRequiredPaymentEvent expected, IncentiveReq
 
             return expected.Type == actual.Type;
         }
+
+        private RequiredPaymentEvent CreateContractTypeRequiredPaymentEvent(decimal amountDue, OnProgrammeEarningType onProgrammeEarningType, byte deliveryPeriod)
+        {
+            ContractType contractType;
+
+            if (currentIlr == null || !currentIlr.Any())
```
samuel-dev-dba: This logic should be moved into the base class as it will be used in Earning Events Matcher

kuzinator: moved to EnumHelper

PR #154, opened 2019-01-29
PV2-588 add step: Only the following provider payments

carlbm:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndSteps.cs
```diff
@@ -220,6 +220,12 @@ public async Task ThenNoPaymentsWillBeCalculated()
             await WaitForUnexpected(() => matcher.MatchNoPayments(), "Required Payment event check failure");
         }
 
+        [Then(@"and only the following provider payments will be generated")]
+        public async Task ThenOnlyTheFolloingProviderPaymentsWillBeGenerated(Table table)
```
davehowlett: typo - should be Following

PR #156, opened 2019-01-30
added ACT1 mapping

davehowlett:


src/SFA.DAS.Payments.Audit.Application.UnitTests/Mapping/EarningEvent/ApprenticeshipContractType1MappingTests.cs
```diff
@@ -0,0 +1,12 @@
+﻿using SFA.DAS.Payments.EarningEvents.Messages.Events;
```
samuel-dev-dba: There is no test to ensure AgreementId is Mapped

PR #158, opened 2019-01-30
PV2-587-add-step-month-end-payment-calculation

carlbm:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndSteps.cs
```diff
@@ -208,9 +208,14 @@ public async Task ThenTheFollowingLearnerEarningsShouldBeGenerated(Table table)
...
-            var matcher = new RequiredPaymentEventMatcher(TestSession, CurrentCollectionPeriod, expectedPayments);
-            await WaitForIt(() => matcher.MatchPayments(), "Required Payment event check failure");
+            await MatchCalculatedPayments(table);
+        }
+
+        [Then(@"at month end only the following payments will be calculated")]
+        public async Task ThenAtMonthEndOnlyTheFollowingPaymentsWillBeCalculated(Table table)
+        {
+            await StartMonthEnd();
+            await MatchCalculatedPayments(table);
```
samuel-dev-dba: This is not right has Required Payment service has no knowledge of month end.
So no need to send  start month end before validating, this should be pointing to Levy Funding source service

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndStepsBase.cs
```diff
@@ -573,9 +574,23 @@ protected async Task StartMonthEnd()
                     MessageId = monthEndCommand.CommandId
                 }}
             };
-            await MessageSession.Send(startedMonthEndJob).ConfigureAwait(false);
 
-            await MessageSession.Send(monthEndCommand).ConfigureAwait(false);
+            var startMonthEndJob2 = new ProcessLevyPaymentsOnMonthEndCommand
```
kuzinator: Can the name be a bit more descriptive? Other commands here were poorly named, with the third command in the list having virtually the same name it's hard to work out what's going on.

davehowlett: I agree - maybe name them after the actual class they represent?

carlbm: Ah! I forgot to rename this one 😈

carlbm:
ALready in feature branch

PR #167, opened 2019-02-04
PV2-620-datalock-actor-map-earning-event

carlbm:
Datalocks service now returning passthough mapping
Removed readonly list from events (issues with mapping)
Added test project for the datalocks service


src/SFA.DAS.Payments.Datalocks.Messages/DataLockEvent.cs
```diff
@@ -8,11 +8,11 @@ namespace SFA.DAS.Payments.DataLocks.Messages
 {
     public abstract class DataLockEvent : PaymentsEvent, IContractType1EarningEvent
     {
-        public ReadOnlyCollection<PriceEpisode> PriceEpisodes { get; set; }
+        public List<PriceEpisode> PriceEpisodes { get; set; }
```
kuzinator: Whiy change? Why to List?

src/SFA.DAS.Payments.DataLocks.DataLockService/DataLockService.cs
```diff
@@ -11,14 +12,16 @@ namespace SFA.DAS.Payments.DataLocks.DataLockService
...
-        public DataLockService(ActorService actorService, ActorId actorId): base(actorService, actorId)
+        private readonly IMapper mapper;
+
+        public DataLockService(ActorService actorService, ActorId actorId, IMapper mapper) : base(actorService, actorId)
         {
+            this.mapper = mapper;
         }
 
-        public Task<DataLockEvent> HandlePayment(ApprenticeshipContractType1EarningEvent message, CancellationToken cancellationToken)
+        public async Task<DataLockEvent> HandlePayment(ApprenticeshipContractType1EarningEvent message, CancellationToken cancellationToken)
```
davehowlett: Should this method be called HandleEarning?

davehowlett: Can the return value be a PayableEarningEvent? This result is going to be published by the Proxy service and the receiving service expects a PayableEarningEvent, not a DataLockEvent

src/SFA.DAS.Payments.Datalocks.Messages/DataLockEvent.cs
carlbm: Automapper...

src/SFA.DAS.Payments.DataLocks.DataLockService/DataLockService.cs
davehowlett: There is no await in the method. Does this need to be async?

carlbm: So what happens with a non-payable earning?

carlbm: It helps with the casting of the generics. May be irrelevant depending on how we deal with DatalockEvent vs PayableEarningsEvent

davehowlett: good point. At the moment, we just require a PayableEarningEvent, but down the wire the DataLockService will be returning the correct object type that the handler can then just publish (and which will then be routed to the correct place down the line).

Happy to leave it as it is.

davehowlett: interesting - after pulling this branch into PV2-619 the warning about async not being required has disappeared! Happy to leave this

carlbm:

PR #175, opened 2019-02-10
PV2-631 - new step for commitment modifications. Addition of VersionI…

davehowlett:
…d to commitment table.

Needs database deployment


kuzinator:
ok to go after latest changes

PR #181, opened 2019-02-19
Pv2 635 e2e step for capping

carlbm:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/SFA.DAS.Payments.AcceptanceTests.EndToEnd.csproj
```diff
@@ -638,6 +648,14 @@
       <Generator>SpecFlowSingleFileGenerator</Generator>
       <LastGenOutput>Non-Levy learner, provider changes aim sequence numbers PV2-393.feature.cs</LastGenOutput>
     </None>
+    <None Include="Tests\Capping\Levy learner over funding band cap - PV2-528.feature">
```
davehowlett: The only question I have is whether we split the tests into the non-levy and levy folders and then have a Capping folder off that or the way you've done it?

PR #190, opened 2019-02-22
Pv2 296 levy learner planned break

carlbm:
This PR is against the 294/295 branch. Will need to change once that's in master


carlbm:

PR #194, opened 2019-02-25
Pv2 267 2 learners 2 employers 1 provider enough levy

kuzinator:
fix for audit service lag, fix for reliable dictionary initialisation

### DIFF
--- a/src/SFA.DAS.Payments.Audit.Application/PaymentsEventProcessing/PaymentsEventModelBatchService.cs
+++ b/src/SFA.DAS.Payments.Audit.Application/PaymentsEventProcessing/PaymentsEventModelBatchService.cs
@@ -11,6 +11,7 @@ namespace SFA.DAS.Payments.Audit.Application.PaymentsEventProcessing
     public interface IPaymentsEventModelBatchService<T> where T : IPaymentsEventModel
     {
         Task RunAsync(CancellationToken cancellationToken);
+        Task StorePayments(CancellationToken cancellationToken);
     }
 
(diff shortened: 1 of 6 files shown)


src/SFA.DAS.Payments.Audit.Application/PaymentsEventProcessing/PaymentsEventModelBatchService.cs
```diff
@@ -11,6 +11,7 @@ namespace SFA.DAS.Payments.Audit.Application.PaymentsEventProcessing
     public interface IPaymentsEventModelBatchService<T> where T : IPaymentsEventModel
     {
         Task RunAsync(CancellationToken cancellationToken);
+        Task StorePayments(CancellationToken cancellationToken);
```
carlbm: Can we call this Flush?

src/SFA.DAS.Payments.ProviderPayments.ProviderPaymentsService/Cache/MonthEndCache.cs
```diff
@@ -51,13 +52,33 @@ public async Task<MonthEndDetails> GetMonthEndDetails(long ukprn, short academic
 
         private async Task<IReliableDictionary2<string, MonthEndDetails>> GetState()
         {
-            if (reliableDictionary != null) return reliableDictionary;
+            if (reliableDictionary != null)
```
carlbm: Considering that this is coming on the back of a race condition.
And that  the state manage was timing out the lock at 2 seconds...

Can we put this into a thread-safe structure? Lazy works. Or something else, but this makes me uncomfortable

PR #204, opened 2019-02-27
PV2-680_new_end_to_end_step_required_payments_step

kuzinator:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndSteps.cs
```diff
@@ -221,16 +221,17 @@ public async Task ThenNoPaymentsWillBeCalculated()
...
+            await ThenOnlyTheFollowingProviderPaymentsWillBeGenerated(TestSession.Provider.Identifier, table).ConfigureAwait(false);
         }
 
-        [When(@"only the following ""(.*)"" payments will be generated")]
-        public async Task WhenOnlyTheFollowingPaymentsWillBeGenerated(string providerIdentifier, Table table)
+        [Then(@"only the following ""(.*)"" payments will be generated")]
+        public async Task ThenOnlyTheFollowingProviderPaymentsWillBeGenerated(string providerIdentifier, Table table)
         {
             var provider = TestSession.GetProviderByIdentifier(providerIdentifier);
+            await StartMonthEnd(provider).ConfigureAwait(false);
```
davehowlett: Does the conversation with Kola about the new step for month end for multiple providers have any bearing on this?

samuel-dev-dba: Please revert back this step to the previous state as we now have a new When Month End is triggered step. This will be merged when 270 is in master

kuzinator: done

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndSteps.cs
```diff
@@ -232,17 +232,16 @@ public async Task ThenAtMonthEndNoPaymentsWillBeCalculatedForProvider(string pro
         public async Task ThenOnlyTheFollowingProviderPaymentsWillBeGenerated(Table table)
         {
             await StartMonthEnd(TestSession.Provider).ConfigureAwait(false);
-            await MatchOnlyProviderPayments(table, TestSession.Provider).ConfigureAwait(false);
+            await ThenOnlyTheFollowingPaymentsWillBeGenerated(TestSession.Provider.Identifier, table).ConfigureAwait(false);
```
samuel-dev-dba: Why are we changing method? 
We already have the provider there is no need to call a method that will start searching for it using the identifier.

kuzinator: reverted

PR #215, opened 2019-03-06
PV2-702 Rename messages

carlbm:


src/SFA.DAS.Payments.RequiredPayments.Messages/Events/CalculatedRequiredLevyAmount.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace SFA.DAS.Payments.RequiredPayments.Messages.Events
 {
-    public class ApprenticeshipContractType1RequiredPaymentEvent : ApprenticeshipContractTypeRequiredPaymentEvent
+    public class CalculatedRequiredLevyAmount : ApprenticeshipContractTypeRequiredPaymentEvent
```
kuzinator: `ApprenticeshipContractTypeRequiredPaymentEvent` class now looks really odd here.
I Suggest `CalculatedRequiredAmount` or `CalculatedRequiredMainAimAmount`

carlbm: Can that happen in a new story?

PR #219, opened 2019-03-07
Pv2 721 rename not levy funding source handlers

carlbm:


src/SFA.DAS.Payments.FundingSource.NonLevyFundedService/Handlers/CalculatedRequiredIncentiveAmountHandler.cs
```diff
@@ -8,13 +8,13 @@
 
 namespace SFA.DAS.Payments.FundingSource.NonLevyFundedService.Handlers
 {
-    public class IncentiveRequiredPaymentEventHandler : IHandleMessages<CalculatedRequiredIncentiveAmount>
+    public class CalculatedRequiredIncentiveAmountHandler : IHandleMessages<CalculatedRequiredIncentiveAmount>
```
kuzinator: CalculatedRequiredIncentiveAmount**Event**Handler ?

src/SFA.DAS.Payments.FundingSource.NonLevyFundedService/Handlers/CalculatedRequiredIncentiveAmountHandler.cs
```diff
@@ -8,13 +8,13 @@
...
     {
         private readonly IPaymentLogger logger;
         private readonly IExecutionContext executionContext;
         private readonly IIncentiveRequiredPaymentProcessor incentiveProcessor;
```
kuzinator: Should this be renamed too since the message it handles changed name?

src/SFA.DAS.Payments.FundingSource.NonLevyFundedService/Handlers/CalculatedRequiredIncentiveAmountHandler.cs
carlbm: done

src/SFA.DAS.Payments.FundingSource.NonLevyFundedService/Handlers/CalculatedRequiredIncentiveAmountHandler.cs
carlbm: No, it's procesing incentives so the name makes sense

PR #229, opened 2019-03-12
Pv2 759 tech fixes for refund tests

kuzinator:
first commit from from another branch with a separate PR, pls ignore it, it was required for this one though.

### DIFF
--- a/src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndStepsBase.cs
+++ b/src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndStepsBase.cs
@@ -812,6 +812,11 @@ namespace SFA.DAS.Payments.AcceptanceTests.EndToEnd.Steps
                     {
                         foreach (var currentPriceEpisode in CurrentPriceEpisodes)
                         {
+                            if (currentPriceEpisode.SfaContributionPercentage == null)
+                            {
+                                currentPriceEpisode.SfaContributionPercentage = training.SfaContributionPercentage;
+                            }
+
                             if (currentPriceEpisode.AimSequenceNumber == 0)
                             {
                                 aims.Single().PriceEpisodes.Add(currentPriceEpisode);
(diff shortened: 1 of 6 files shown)


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndStepsBase.cs
```diff
@@ -812,6 +812,11 @@ protected async Task GeneratedAndValidateEarnings(Table table, Provider provider
...
                         foreach (var currentPriceEpisode in CurrentPriceEpisodes)
                         {
+                            if (currentPriceEpisode.SfaContributionPercentage == null)
+                            {
+                                currentPriceEpisode.SfaContributionPercentage = training.SfaContributionPercentage;
+                            }
+
                             if (currentPriceEpisode.AimSequenceNumber == 0)
                             {
                                 aims.Single().PriceEpisodes.Add(currentPriceEpisode);
```
davehowlett: Not your code but a bit smelly - there is only one Aim at this point, so why not use that?

carlbm: The check with single is good to stop this path happening when the spec doesn't expect it?

kuzinator: When there is no AimSequenceNumber specified, there has to be just one aim, this Single makes sure of it.

davehowlett: I understand that, but if you look at where aim and aims are declared at the top of the foreach loop:

`  foreach (var training in providerCurrentIlrs)
                {
                    var aim = new Aim(training);
                    var aims = new List<Aim> { aim };
                    AddTestAims(aims, provider.Ukprn);`

you can see that "aim" is a singular item and "aims" has only been created so that the aim can be passed to the AddTestAims method

kuzinator: changed

PR #231, opened 2019-03-13
Feature/refunds payments due v2

carlbm:


carlbm:
This is just here to check that it's up to date with master

PR #245, opened 2019-03-21
Pv2 741 providerpayment correct act

davehowlett:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Tests/Learner Change of Circumstances/Levy learner changes price during month - PV2-295.feature
```diff
@@ -30,9 +30,9 @@ Scenario Outline: Levy learner changes to the agreed price during the month PV2-
 	# Additional fields
 	# Do I need to move the second commitment after history?	
 	And the following commitments exist
-        | commitment Id | version Id | start date                   | end date                  | status | agreed price | effective from               | effective to                 |
-        | 1             | 1-001      | 01/Aug/Current Academic Year | 31/Aug/Next Academic Year | active | 15000        | 01/Aug/Current Academic Year | 31/Oct/Current Academic Year |
-        | 1             | 1-002      | 01/Aug/Current Academic Year | 31/Aug/Next Academic Year | active | 9375         | 01/Nov/Current Academic Year |                              |
+        | start date                   | end date                  | status | agreed price | effective from               | effective to                 |
```
carlbm: The commitment id is the same (ie relevant), so what we've done elsewhere is to use the story number as the commitment id. You'll need to leave both the commitment id and version id columns in though

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/EventMatchers/ProviderPaymentModelMatcher.cs
```diff
@@ -50,31 +51,38 @@ protected override IList<PaymentModel> GetExpectedEvents()
 
             foreach (var paymentInfo in expectedPaymentInfo)
             {
+                var learner = testSession.GetLearner(provider.Ukprn, paymentInfo.LearnerId);
+
+                var currentIlrForLearner =
+                    providerCurrentIlr?.Where(x => x.LearnerId == paymentInfo.LearnerId).ToList();
+                
+                var contractType = GetContractType(currentIlrForLearner, learner, paymentInfo.TransactionType);
```
carlbm: I have an issue with this. The contract type is a price episode property so using the entire ILR seems like we're setting ourselves up for a fall when we need to deal with it (in fact how does this play out with Kola's exisitng PR?). I would prefer that we put the contract type wherever it is relevant in the tests and get rid of all the code that infers it

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Tests/Learner Change of Circumstances/Levy learner changes price during month - PV2-295.feature
davehowlett: ok - when I ran the tests with what had been in master before, this test failed. I'll add them back in as you suggested and pull master again

davehowlett:
to be addressed differently

PR #247, opened 2019-03-21
Release/release pv2 252 pv2 770 pv2 774 pv2 639 pv2 767

Ron2404:


Ron2404:
This build has passed in Das Test after ignoring test for PV2-252 
Build id : 1816 
Release id : DAS Payments V2-0.0.1816-R001

PR #261, opened 2019-03-27
Pv2 801 apprenticeship key factory

kuzinator:


src/SFA.DAS.Payments.ServiceFabric.Core/Infrastructure/Ioc/ServiceFabricContainerFactory.cs
```diff
@@ -13,7 +13,11 @@ public static class ServiceFabricContainerFactory
         {
             var builder = ContainerFactory.CreateBuilder();
             builder.RegisterActor<TActor>()
-                .OnActivating(e => ((ActorStateManagerProvider) e.Context.Resolve<IActorStateManagerProvider>()).Current = e.Instance.StateManager);            
+                .OnActivating(e =>
+                {
+                    ((ActorStateManagerProvider) e.Context.Resolve<IActorStateManagerProvider>()).Current = e.Instance.StateManager;
```
carlbm: This seems fragile. Why not resolve `ActorStateManager`?

PR #262, opened 2019-03-28
PV2-797 add completion payment held back event

carlbm:


src/SFA.DAS.Payments.EarningEvents.Application/Handlers/ProcessLearnerHandler.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
```
kuzinator: Is that significant?

PR #267, opened 2019-04-01
Fix Not Logging bug - PV2-785

ESFA-Pawan:


ESFA-Pawan:
Passed in DAS - DAS Payments V2-0.0.1868-R001

PR #276, opened 2019-04-03
Pv2 342 add earnings info to provider payment events

davehowlett:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndStepsBase.cs
```diff
@@ -310,7 +310,14 @@ private PaymentModel CreatePaymentModel(ProviderPayment providerPayment, Trainin
                 LearnerUln = providerPayment.Uln,
                 LearningAimFrameworkCode = learnerTraining.FrameworkCode,
                 LearningAimProgrammeType = learnerTraining.ProgrammeType,
-                AccountId = accountId
+                AccountId = accountId,
+                StartDate = DateTime.UtcNow,
+                PlannedEndDate = DateTime.UtcNow,
+                ActualEndDate = DateTime.UtcNow,
+                CompletionStatus = 1,
```
samuel-dev-dba: We shouldn't have hardcoded values in test, they should come from spec

davehowlett: they don't have a place in the spec - once the mapping is done from Earning Events, then this can be done. I raised PV2-859 to resolve that issue

PR #285, opened 2019-04-05
Pv2 763 payable earning application service

davehowlett:


src/SFA.DAS.Payments.DataLocks.Domain/Interfaces/ILearnerMatcher.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Threading.Tasks;
+using SFA.DAS.Payments.DataLocks.Domain.Models;
+
+namespace SFA.DAS.Payments.DataLocks.Domain.Interfaces
+{
+    public interface ILearnerMatcher
+    {
+        Task<LearnerMatchResult> MatchLearner(long ukprn, long uln);
```
samuel-dev-dba: Please remove the ukprn parameter as the actor is already scoped to Ukprn so not needed.

davehowlett: removed

src/SFA.DAS.Payments.DataLocks.Application/Infrastructure/ioc/DataLockModule.cs
```diff
@@ -16,7 +16,9 @@ protected override void Load(ContainerBuilder builder)
             builder.RegisterType<ActorReliableCollectionCache<List<ApprenticeshipModel>>>().AsImplementedInterfaces().InstancePerLifetimeScope();
             builder.RegisterType<DataLockLearnerCache>().As<IDataLockLearnerCache>().InstancePerLifetimeScope();
             builder.RegisterType<UkprnMatcher>().As<IUkprnMatcher>().InstancePerLifetimeScope();
-            builder.RegisterType<UlnLearnerMatcher>().AsImplementedInterfaces().InstancePerLifetimeScope();
+            builder.RegisterType<UlnLearnerMatcher>().As<IUlnLearnerMatcher>().InstancePerLifetimeScope();
```
kuzinator: This looks like a merge issue, should use AsImplementedInterfaces

src/SFA.DAS.Payments.DataLocks.DataLockService/DataLockService.cs
```diff
@@ -41,15 +46,34 @@ public class DataLockService : Actor, IDataLockService
...
+            this.learnerMatcher = learnerMatcher;
+            this.courseValidator = courseValidator;
         }
 
         public async Task<DataLockEvent> HandleEarning(ApprenticeshipContractType1EarningEvent message, CancellationToken cancellationToken)
         {
-            var apprenticeshipsForUln = await apprenticeships.TryGet(message.Learner.Uln.ToString(), cancellationToken)
-                .ConfigureAwait(false);
-            var apprenticeship = apprenticeshipsForUln.Value.FirstOrDefault();
+            var learnerMatchResult = await learnerMatcher.MatchLearner(message.Ukprn);
```
kuzinator: I think we have problems with naming in our design. DataLockService is an actor, i.e. service layer, on our design we have DataLockService which is app layer and is NOT the same service.

src/SFA.DAS.Payments.DataLocks.DataLockService/DataLockService.cs
```diff
@@ -41,15 +46,34 @@ public class DataLockService : Actor, IDataLockService
...
+
+            if (learnerMatchResult.DataLockErrorCode.HasValue)
+            {
+                // TODO: return non-payable earning
+                return null;
```
benagain: I don't think we should return `null` from a `Task<T>` method, it'll cause a NullReferenceException in the calling `await` without the caller having to do some null checking.  

```suggestion
                return Task.FromResult<DataLockEvent>(null);
```

src/SFA.DAS.Payments.DataLocks.DataLockService/DataLockService.cs
```diff
@@ -41,15 +46,34 @@ public class DataLockService : Actor, IDataLockService
...
 
             var returnMessage = mapper.Map<PayableEarningEvent>(message);
+
+            // TODO: After implementing CourseValidator, need correct message going into CourseValidator
+            var courseValidationResult = await courseValidator.ValidateCourse(returnMessage.CollectionPeriod, apprenticeshipsForUln);
+
+            if (courseValidationResult.ValidationResults.Any(x => x.DataLockErrorCode.HasValue))
+            {
+                // TODO: return non-payable earning
+                return null;
```
benagain: Same as above

```suggestion
                return Task.FromResult<DataLockEvent>(null);
```

src/SFA.DAS.Payments.DataLocks.DataLockService/DataLockService.cs
samuel-dev-dba: Input to learnerMatcher is Uln NOT ukprn

src/SFA.DAS.Payments.DataLocks.DataLockService/DataLockService.cs
davehowlett: This is only a temporary thing. one of the DataLock component tasks involves publishing NonPayableEvents and at that point, the correct values could be returned.

Adding the suggested code in doesn't compile as the method is async so the result needs to be a DataLockEvent.

src/SFA.DAS.Payments.DataLocks.DataLockService/DataLockService.cs
davehowlett: done - new DataLockProcessor added and all logic in the DataLockService Actor has been moved to the processor.

src/SFA.DAS.Payments.DataLocks.DataLockService/DataLockService.cs
benagain: Ah yes, I missed the `async` on the method.

src/SFA.DAS.Payments.DataLocks.DataLockService/DataLockService.cs
davehowlett: done

src/SFA.DAS.Payments.DataLocks.Application/Services/DataLockProcessor.cs
```diff
@@ -0,0 +1,66 @@
...
+                // TODO: return non-payable earning
+                return null;
+            }
+
+            var apprenticeshipsForUln = learnerMatchResult.Apprenticeships;
+
+            var courseValidationResult = await courseValidator.ValidateCourse(courseValidation, apprenticeshipsForUln);
+
+            if (courseValidationResult.ValidationResults.Any(x => x.DataLockErrorCode.HasValue))
+            {
```
samuel-dev-dba: This logic is not right. We don't stop payment because one period failed datalock.

src/SFA.DAS.Payments.DataLocks.Domain/Models/CourseValidationResult.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.Generic;
+
+namespace SFA.DAS.Payments.DataLocks.Domain.Models
+{
+    public class CourseValidationResult
+    {
+        public List<ValidationResult> ValidationResults { get; set; }
```
samuel-dev-dba: Can we just return List of ValidationResults  I don't think we need a class here

PR #301, opened 2019-04-16
Pv2 835 publishing nonpayable earnings

davehowlett:


src/SFA.DAS.Payments.DataLocks.Application/Services/DataLockProcessor.cs
```diff
@@ -27,64 +29,115 @@ public DataLockProcessor(IMapper mapper, ILearnerMatcher learnerMatcher, IProces
...
+                var onProgrammeEarningPeriods = ValidateOnProgEarningPeriods(earningEvent.Learner.Uln, earningEvent.PriceEpisodes, onProgrammeEarning, apprenticeshipsForUln);
 
-            var courseValidationResult = await processCourseValidator.ValidateCourse(courseValidation, apprenticeshipsForUln);
+                var validPeriods = onProgrammeEarningPeriods.ValidPeriods;
+                var validationErrors = onProgrammeEarningPeriods.ValidationErrors;
 
-            var returnMessage = mapper.Map<PayableEarningEvent>(earningEvent);
+                if (validationErrors != null && validationErrors.Any())
+                {
+                    result.Add(CreateNonPayableEarningEvent(earningEvent, validationErrors));
```
samuel-dev-dba: This is not right we should bring back only the periods that fail Datalock and then create a Non-Payable DatalockEvent for then. This logic implies that all periods fail DataLock which is not always right.

davehowlett:
different implementation being done by Kola

PR #303, opened 2019-04-16
PV2-878 issues with specs

carlbm:


carlbm:
Includes story PV2-879 too

PR #313, opened 2019-04-20
Pv2 913 557 and english and maths

carlbm:
This includes all english and maths stories
285, 284, 287, 288 are ignored for now as they rely on bad provider behaviour module

All tests pass locally


src/SFA.DAS.Payments.RequiredPayments.Application.UnitTests/Application/EarningEventMappingTest.cs
```diff
@@ -116,6 +113,22 @@ public void ContractTypeIsCorrectForNotLevyEvent(Type earningEventType, Type req
...
+        [Test]
+        [TestCase(typeof(CalculatedRequiredIncentiveAmount))]
+        [TestCase(typeof(CalculatedRequiredCoInvestedAmount))]
+        [TestCase(typeof(CalculatedRequiredLevyAmount))]
+        public void ContractTypeIsCorrectForFunctionalSkills(Type requiredPaymentEventType)
+        {
+            var requiredPaymentEvent = Activator.CreateInstance(requiredPaymentEventType) as RequiredPaymentEvent;
+            var earningEvent = new FunctionalSkillEarningsEvent
+            {
+                ContractType = ContractType.Act1,
```
abroodie: If you put the expected contract type as a parameter to the method and test case then this can test the mapping for ACT1 & ACT2

src/SFA.DAS.Payments.AcceptanceTests.Core/AimPeriodMatcher.cs
```diff
@@ -20,85 +16,7 @@ public static class AimPeriodMatcher
...
-                 aimStartDate <= collectionPeriodReferenceDate &&
-                 DurationGreaterThanCollectionPeriodStart(aimEndDate, collectionPeriodReferenceDate)))
-            {
-                return true;
-            }
-
-            return aimStartPeriod.AcademicYear == collectionPeriod.AcademicYear &&
-                   aimStartDate <= collectionPeriodReferenceDate &&
-                   DurationGreaterThanCollectionPeriodStart(aimEndDate, collectionPeriodReferenceDate);
+            return true;
```
abroodie: this seems like a fairly big and unexpected change.  I think we need to discuss this.

PR #324, opened 2019-04-26
Pv2 884 lookahead for inconsistent submission

kuzinator:
Inconsistent submission tests are added to the solution and named correctly.
Small fixes done to features to include SFA contribution % for maths & english aims
Handling of Learn Aim Reference in earnings table


ESFA-Pawan:
@kuzinator PV2-464, not sure if earning calc will return -ve value for maths/eng aim but happy to accept the change. We will not at run-time.

But the change to the start date is not right. Start date shouldn't have been changed to Oct in 'But aims details are changed as follows' section. Please revert.

PR #358, opened 2019-05-14
Release/release pv2 977 reply functionality

Ron2404:


Ron2404:
DAS Payments V2-0.0.2150-R001  Build :0.0.2150  passed in DAS-Test

PR #366, opened 2019-05-16
PV2-941 Send JobContextMessage instead of ProcessLearnerCommand

carlbm:
…sage) instead of ProcessLearnerCommand


carlbm:
It's important to check that the DcStorageConnectionString configuration is the same between the earnings service and the test harness

PR #368, opened 2019-05-16
Pv2 921 removed learning aim service

kuzinator:


src/SFA.DAS.Payments.Application/Repositories/ISubmittedLearnerAimRepository.cs
```diff
@@ -19,16 +19,18 @@ public SubmittedLearnerAimRepository(IPaymentsDataContext paymentsDataContext)
...
         {
             return await paymentsDataContext.Database.ExecuteSqlCommandAsync(@"
                 delete from [Payments2].[SubmittedLearnerAim] 
                 where 
-                    [AcademicYear] = @academicYear 
-                    and [CollectionPeriod] = @period 
-                    and [IlrSubmissionDateTime] <= @newIlrSubmissionDate",
-                cancellationToken,
-                new {academicYear, period, newIlrSubmissionDate}).ConfigureAwait(false);
+                    [Ukprn] = {0}
```
carlbm: why not use named parameters?
Or use the $ syntax

src/SFA.DAS.Payments.RequiredPayments.RequiredPaymentsProxyService/Handlers/ReceivedProviderEarningsEventHandler.cs
```diff
@@ -31,6 +31,7 @@ public async Task Handle(ReceivedProviderEarningsEvent message, IMessageHandlerC
 
             foreach (var removedAim in removedAims)
             {
+                removedAim.JobId = message.JobId;
                 await context.Publish(removedAim).ConfigureAwait(false);
```
carlbm: Not sure how many messages we're dealing with here, but does it make sense to add all the tasks to a container and then `Task.WhenAll()` on the container?

PR #369, opened 2019-05-20
Release/release20052019

Ron2404:
This contains the following tickets :
PV2-921, PV2-917,PV2-932, PV2-464,PV2-941 .


samuel-dev-dba:
Not needed by testers anymore

PR #399, opened 2019-06-03
Pv2 1025 add required reporting fields to the payments table

abroodie:


src/SFA.DAS.Payments.FundingSource.Application.UnitTests/Mapper/CalculatedRequiredLevyAmountMapperTest.cs
```diff
@@ -56,16 +60,21 @@ public void Setup()
             autoMapper = mapperConfiguration.CreateMapper();
         }
 
+        protected void CompareCommonProperties(CalculatedRequiredLevyAmount source, FundingSourcePaymentEvent destination, List<string> propertiesToIgnore)
```
samuel-dev-dba: I will have prefered that we match each property rather than using reflection to match them it makes it clearer to see what we are testing.

abroodie: done

PR #408, opened 2019-06-06
Pv2 1084 remove last year price episodes from earnings

kuzinator:


src/SFA.DAS.Payments.EarningEvents.Application.UnitTests/Mapping/MappingTest.cs
```diff
@@ -561,6 +564,206 @@ public void TestMultiplePriceEpisodes()
...
+                    TNP1 = 7500,
+                    TNP2 = 7500,
+                    PriceEpisodeCompleted = true,
+                },
+                PriceEpisodePeriodisedValues = new List<PriceEpisodePeriodisedValues>
+                {
+                    new PriceEpisodePeriodisedValues
+                    {
+                        AttributeName = "PriceEpisodeOnProgPayment",
+                        Period1 = 500,
```
carlbm: Last year (and future year) prices will always be 0

src/SFA.DAS.Payments.EarningEvents.Application.UnitTests/Mapping/MappingTest.cs
```diff
@@ -561,6 +564,206 @@ public void TestMultiplePriceEpisodes()
...
+                        Period8 = 0,
+                        Period9 = 0,
+                        Period10 = 0,
+                        Period11 = 0,
+                        Period12 = 0
+                    },
+                    new PriceEpisodePeriodisedValues
+                    {
+                        AttributeName = "PriceEpisodeCompletionPayment",
+                        Period1 = 100,
```
carlbm: Last year (and future year) prices will always be 0

src/SFA.DAS.Payments.EarningEvents.Application/Mapping/IntermediateLearningAim.cs
```diff
@@ -43,14 +43,35 @@ public class IntermediateLearningAim
...
             Ukprn = ukprn;
             AcademicYear = collectionYear;
             CollectionPeriod = collectionPeriod;
             IlrSubmissionDateTime = ilrSubmissionDateTime;
             JobId = jobId;
         }
 
+        private IEnumerable<PriceEpisode> FilterPriceEpisodes(IEnumerable<PriceEpisode> priceEpisodes, short currentAcademicYear)
+        {
+            var calendarYear = currentAcademicYear / 100 + 2000;
```
carlbm: What happens in the 9900 academic year?

src/SFA.DAS.Payments.EarningEvents.Application.UnitTests/Mapping/MappingTest.cs
```diff
@@ -561,6 +564,206 @@ public void TestMultiplePriceEpisodes()
...
+                        Period3 = 0,
+                        Period4 = 0,
+                        Period5 = 0,
+                        Period6 = 0,
+                        Period7 = 0,
+                        Period8 = 0,
+                        Period9 = 0,
+                        Period10 = 0,
+                        Period11 = 0,
+                        Period12 = 500
```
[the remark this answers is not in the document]
kuzinator: noted

src/SFA.DAS.Payments.EarningEvents.Application/Mapping/IntermediateLearningAim.cs
kuzinator: Let's wait 80 years and see

samuel-dev-dba: I think is about time we have an academic year dates service so we can have this logic in one place seems to be everywhere now.

kuzinator: I'd love to bring back CalendarPeriod class with all logic in one place. Separate service to divide and then add - no.

samuel-dev-dba: We had to create one in commitment and it was very useful. The output will be acaemic year start date and end date. The number of line of codes is irrelevant here, what matters is that we have something that handles a business logic in a central place.

kuzinator: Will consider that when we have one more place in production code that does the same. This particular logic is not used anywhere else apart from test.

samuel-dev-dba: Ok fair enough

PR #409, opened 2019-06-06
Release/release pv2 1017 pv2 256

ESFA-Pawan:


ESFA-Pawan:
Release passed in both DASTEST and DSI but for unknown reason there are 5 tests for PV2-256 instead of 2. 2 tests are running fine and 3 are ignored. I'll raise a separate ticket for this observation.

PR #413, opened 2019-06-07
Pv2 1072 create a web endpoint to trigger month end events for manual testing

samuel-dev-dba:


src/SFA.DAS.Payments.PeriodEnd.TestEndpoint/Infrastructure/EndpointNames.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace SFA.DAS.Payments.PeriodEnd.TestEndpoint.Infrastructure
+{
+    public class EndpointNames
+    {
+        public const string ProviderPayments = "sfa-das-payments-providerpayments";
```
carlbm: Shouldn't this be a config setting?

src/SFA.DAS.Payments.PeriodEnd.TestEndpoint/Properties/launchSettings.json
```diff
@@ -0,0 +1,27 @@
+﻿{
+  "iisSettings": {
+    "windowsAuthentication": false, 
+    "anonymousAuthentication": true, 
+    "iisExpress": {
+      "applicationUrl": "http://localhost:6738",
```
carlbm: Shouldn't the port be a config setting?
Also, not sure that localhost will work? Maybe 0.0.0.0 if it doesn't

samuel-dev-dba: This only applies to the dev environment

src/SFA.DAS.Payments.PeriodEnd.TestEndpoint/Infrastructure/EndpointNames.cs
samuel-dev-dba: done

samuel-dev-dba:
No Need to merge to master will be manually deployed when required.

PR #425, opened 2019-06-11
Pv2 1003 fix performance tests project

abroodie:


src/SFA.DAS.Payments.PerformanceTests/EndToEndPerformanceTests.cs
```diff
@@ -1,146 +1,246 @@
...
+
+        protected DateTimeOffset? DeliveryTime;
+
+        protected async Task SubmitIlr(TestSession session, int collectionPeriod)
+        {
+            var ilrLearners = session.Learners
+                .Select(learner => CreateFM36Learner(session, learner))
+                .ToList();
+            session.IlrSubmissionTime = DateTime.UtcNow;
+            //foreach (var fm36Learner in ilrLearners)
```
carlbm: If we don't need the commented code can you remove it?

abroodie: done

PR #428, opened 2019-06-13
PV2-1155: audit service improvements

kuzinator:
- better logging
- do not crash on first exception
- try one record by one if batch failed and record failing one to log


src/SFA.DAS.Payments.Audit.Application/PaymentsEventProcessing/PaymentsEventModelBatchProcessor.cs
```diff
@@ -50,43 +51,111 @@ public async Task<int> Process(int batchSize, CancellationToken cancellationToke
                 logger.LogVerbose("No records found to process.");
                 return 0;
             }
+
             logger.LogDebug($"Processing {batch.Count} records.");
             var data = dataTable.GetDataTable(batch);
             using (var sqlConnection = new SqlConnection(connectionString))
+            using (var scope = new TransactionScope(TransactionScopeOption.RequiresNew, TransactionScopeAsyncFlowOption.Enabled))
```
carlbm: Not sure that this transaction is required as there is already a transaction at this point. This transaction may even interfere if there is a problem outside of this transaction but within the outer transaction

kuzinator: this is outside of scope of this PR but I agree, this looks like a nested transaction that doesn't bring much value.

PR #435, opened 2019-06-17
PV2-1168 Added the new step: 'the provider prority order is'

carlbm:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndSteps.cs
```diff
@@ -127,6 +128,48 @@ public void GivenPriceDetailsAreChangedAsFollows(Table table)
             GivenPriceDetailsAsFollows(table);
         }
 
+        [Given(@"the provider prority order is")]
+        public void GivenTheProviderPriorityOrder(Table table)
+        {
```
samuel-dev-dba: Can this logic be moved to a helper class that handles employer provider priorities?

PR #447, opened 2019-06-19
PDI-32 PDI-31 small employer EEF codes

CarlHA:
PDI 31 - The funding line type needs to stay as 16-18 as this is what comes out of the FM36 file and is part of the matching criteria.  The DOB is overwritten to be 19+ when generating the ILR.


CarlHA:
Replaced by #457

PR #470, opened 2019-06-26
PDI-42 Changes for 390

DaveGill-Nimbus:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/LearnerMutators/LearnerExtensions.cs
```diff
@@ -22,18 +22,18 @@ public static int ToLearnerAge(this string fundingLineType)
             }
         }
 
-        public static int AsPercentage(this string sfaContributionPercentage)
+        public static int? AsPercentage(this string sfaContributionPercentage)
```
kuzinator: this is not just for SFA% now, can `sfaContributionPercentage`  be renamed?

PR #473, opened 2019-06-27
PV2-1170: changed provider payments to no employer which made test work

davehowlett:


davehowlett:
awaiting comments from @rohannpradhan2404 & @ESFA-Pawan about removing column. Not sure whether there would be an issue in code with removing it and it is correct as per other specs.

samuel-dev-dba:
Closing obsolete  PR

PR #477, opened 2019-06-27
PDI-102 Changes for 526

DaveGill-Nimbus:


CarlHA:
This relies on code from #479

PR #488, opened 2019-07-02
PDI-118 Changes for 501

DaveGill-Nimbus:


CarlHA:
This relies on code from #479 and #487

PR #492, opened 2019-07-03
PDI-120 Changes for 514

DaveGill-Nimbus:


CarlHA:
This relies on code in #479

PR #499, opened 2019-07-04
Pv2 1142 change in contract type description for learning delivery

davehowlett:
branch created from PV2-1167 Maths & English change as well as PV2-1141 price episode contract type description change. This must be merged after those.


src/SFA.DAS.Payments.EarningEvents.Application/Mapping/MappingExtensions.cs
```diff
@@ -1,23 +1,72 @@
...
     {
         private static readonly TypeAccessor PeriodAccessor = TypeAccessor.Create(typeof(PeriodisedAttribute));
+        private static readonly TypeAccessor PeriodTextAccessor = TypeAccessor.Create(typeof(LearningDeliveryPeriodisedTextValues));
 
         public static decimal? GetPeriodValue(this PeriodisedAttribute periodisedValues, int period)
         {
             return (decimal?)PeriodAccessor[periodisedValues, "Period" + period];
         }
 
+        public static string GetPeriodTextValue(this LearningDeliveryPeriodisedTextValues periodisedValues, int period)
```
samuel-dev-dba: should be private  only used within the class

src/SFA.DAS.Payments.EarningEvents.Application/Mapping/MappingExtensions.cs
```diff
@@ -1,23 +1,72 @@
...
+
+        public static ContractType[] GetContractTypesForLearningDeliveries(this List<LearningDelivery> learningDeliveries)
+        {
+            var periodisedTextValues = learningDeliveries.Where(x => x.LearningDeliveryPeriodisedTextValues != null)
+                .SelectMany(x =>
+                    x.LearningDeliveryPeriodisedTextValues.Where(l => l.AttributeName == "LearnDelContType"));
+
+            if (!periodisedTextValues.Any())
+            {
+                return new ContractType[0];
```
samuel-dev-dba: I think we should throw an exception here as we can't have payment with no contract type.

PR #520, opened 2019-07-09
PV2-1278 allow none in contract type

carlbm:
Logic changed so that if it's not ACT1 then it must be ACT2, even if ……it's 'None'


ESFA-Pawan:
Passed in DPT - 
DAS Payments V2-0.0.2727-R001

PR #526, opened 2019-07-10
PDI-42 Changes for 390

DaveGill-Nimbus:


src/SFA.DAS.Payments.AcceptanceTests.Core/Data/Aim.cs
```diff
@@ -77,10 +77,21 @@ CompletionStatus CompletionStatusFromString(string completionStatus)
 
         public bool IsMainAim => AimReference == "ZPROG001";
 
-        public bool IsMathsAndEnglish => AimReference == "50114979" || AimReference == "50093186";
+        public bool IsMathsAndEnglish => MathsAndEnglishAimReferenceCodes.Contains(AimReference);
```
CarlHA: 50114979 is missing from the list of reference codes which will cause 387 to fail

CarlHA: Has now been added when doing updates to 387

PR #532, opened 2019-07-11
Pv2 1288 database updates

carlbm:


src/SFA.DAS.Payments.Database/Payments2/Tables/Payments2.ApprenticeshipPause.sql
```diff
@@ -0,0 +1,11 @@
...
+(
+	[Id] BIGINT IDENTITY(1,1) NOT NULL PRIMARY KEY,
+	ApprenticeshipId BIGINT NOT NULL,
+	PauseDate DATE NOT NULL,
+	ResumeDate DATE NULL
+)
+
+GO
+
+CREATE INDEX [IX_ApprenticeshipPause_ApprenticeshipId] ON [Payments2].[ApprenticeshipPause] (ApprenticeshipId)
```
samuel-dev-dba: Should be a foreign key to Apprenticeships Table

abroodie: Is there a reason this is an index rather than a foreign key to Apprenticeship?

PR #553, opened 2019-07-17
Pv2 1304 bug with payments

carlbm:


carlbm:
Too far out of date so starting again

PR #561, opened 2019-07-18
Pv2 1312 pv2 1320 pv2 1321 functional skills contract types

davehowlett:
Initial PR covering changes to mapping to allow ACT1 and ACT2 functional skills to exist. To be merged to a feature branch as it'll break master


src/SFA.DAS.Payments.RequiredPayments.RequiredPaymentsProxyService/Handlers/EarningEventHandlerBase.cs
```diff
@@ -45,7 +45,7 @@ public async Task Handle(T message, IMessageHandlerContext context)
             {
                 var contractType = message is PayableEarningEvent ? ContractType.Act1 :
                     message is ApprenticeshipContractType2EarningEvent ? ContractType.Act2 :
-                    message is FunctionalSkillEarningsEvent ? (message as FunctionalSkillEarningsEvent).ContractType :
+                    message is ApprenticeshipContractType2FunctionalSkillEarningsEvent ? (message as ApprenticeshipContractType2FunctionalSkillEarningsEvent).ContractType :
```
samuel-dev-dba: Why can't we just return contract type as ACT2

src/SFA.DAS.Payments.RequiredPayments.RequiredPaymentsProxyService/Handlers/FunctionalSkillEarningsEventHandler.cs
```diff
@@ -11,14 +11,14 @@
 
 namespace SFA.DAS.Payments.RequiredPayments.RequiredPaymentsProxyService.Handlers
 {
-    public class FunctionalSkillEarningsEventHandler : EarningEventHandlerBase<FunctionalSkillEarningsEvent>
+    public class FunctionalSkillEarningsEventHandler : EarningEventHandlerBase<ApprenticeshipContractType2FunctionalSkillEarningsEvent>
```
samuel-dev-dba: The class and file name should match the Event that we are handling in the handler.

src/SFA.DAS.Payments.RequiredPayments.RequiredPaymentsProxyService/Handlers/EarningEventHandlerBase.cs
davehowlett: when the datalock code is implemented in one of the other tasks, this will be changed to the base class. See the point, so have changed it to the base class.

src/SFA.DAS.Payments.RequiredPayments.RequiredPaymentsProxyService/Handlers/FunctionalSkillEarningsEventHandler.cs
davehowlett: renamed

PR #564, opened 2019-07-19
Pv2 1324 submission event data tables

kuzinator:
PR is against data lock events feature branch because it has Legacy database in it which is not in master yet


MichaelYoung1981:
Story moving back to analysis so we can determine the correct need.

PR #578, opened 2019-07-25
Pdi 132 levy learner finishes on time pv2 261

keenesolutions:
Updated RemoveEmptyFunctionSkillEarningEven as we were getting null Periods in actualPayment Earnings.


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/EventMatchers/BaseMatcher.cs
```diff
@@ -65,7 +65,7 @@ private IList<T> RemoveEmptyFunctionSkillEarningEvent(IList<T> actualPayments)
         {
             return actualPayments.Except(actualPayments
                 .Where(actualPayment => actualPayment.GetType() == typeof(FunctionalSkillEarningsEvent) && 
-                                          (actualPayment as FunctionalSkillEarningsEvent).Earnings.All(e=>e.Periods.All(p=>p.Amount == 0)))).ToList();
+                                          (actualPayment as FunctionalSkillEarningsEvent).Earnings.All(e=> e.Periods == null  || e.Periods.All(p=>p == null)|| e.Periods.All(p=>p.Amount == 0)))).ToList();
         }
```
CarlHA: Check with Max, but I don't think we would have null periods as an FM36 will output a value for a period even if that is 0.  We may need to have details added to the feature file.

CarlHA: My intended point was that we need to check with Max that this is a valid change and not a symptom of a bug.

CarlHA: Confirmed by Max that this change is Ok

PR #590, opened 2019-07-28
Pv2 1348 disable console logger when not running in dev mode

abroodie:


davehowlett:
Needs config creating so that service fabric build is successful. Code changes look fine.

PR #602, opened 2019-07-31
Month end can't be reversed, so as soon as it's 'triggered' then have…

carlbm:
… a local in-memory cache to check prior to the SF cache check


kuzinator:
This is irrelevant now since there will be no FundingSourceEvents coming after period end event

PR #604, opened 2019-08-01
Pv2 1364 update the monitoring application to use a different database connection string

abroodie:


DavidBSteele:
Not needed anymore

PR #615, opened 2019-08-12
Pv2 1346 update is levy payer flag for levy account and apprenticeships

samuel-dev-dba:


src/SFA.DAS.Payments.DataLocks.Application.UnitTets/Mapping/ApprovalsApprenticeshipEventsMappingTests.cs
```diff
@@ -217,6 +218,28 @@ public void Maps_DataLockTriageApprovedEvent_To_UpdatedApprenticeshipDataLockTri
                 .All(pe => model.ApprenticeshipPriceEpisodes.Any(ape => ape.StartDate == pe.FromDate && ape.Cost == pe.Cost && ape.EndDate == pe.ToDate))
                 .Should().BeTrue();
         }
+        
+        [Test]
+        public void Maps_IsLevyPayer_From_ApprenticeshipUpdated_To_ApprenticeshipModel_Correctly()
```
davehowlett: these two tests could be generalised into a single test with the events passed in

src/SFA.DAS.Payments.DataLocks.Application/Services/ApprenticeshipProcessor.cs
```diff
@@ -212,7 +213,35 @@ public async Task ProcessPaymentOrderChange(PaymentOrderChangedEvent paymentOrde
...
         }
-        
+
+        public async Task ProcessApprenticeshipForNonLevyPayerEmployer(long accountId)
+        {
+            try
+            {
+                logger.LogDebug($"Processing the apprenticeship Employer Is Non Levy Payer event for Account id: {accountId}");
+               
+                var updatedApprenticeships = await apprenticeshipService.UpdateApprenticeshipEmployerIsLevyPayerFlag(accountId).ConfigureAwait(false);
```
davehowlett: could this be better named something like "GetLatestApprenticeshipEmployerIsLevyPayerFlag"? it isn't really clear from it being named Update that it is in fact calling something else to get it and then update locally

src/SFA.DAS.Payments.DataLocks.Domain/Services/Apprenticeships/ApprenticeshipService.cs
```diff
@@ -66,6 +69,22 @@ public async Task<List<ApprenticeshipDuplicateModel>> NewApprenticeship(Apprenti
...
+        public async Task<List<ApprenticeshipModel>> UpdateApprenticeshipEmployerIsLevyPayerFlag(long accountId,CancellationToken cancellation = default(CancellationToken))
+        {
+            var apprenticeships = await repository.GetEmployerApprenticeships(accountId, cancellation).ConfigureAwait(false);
+
+            if (apprenticeships == null || apprenticeships.Count == 0)
+            {
+                return  new List<ApprenticeshipModel>();
+            }
+
+            apprenticeships.ForEach(x => x.IsLevyPayer = false);
```
davehowlett: are they all supposed to be set to false?

src/SFA.DAS.Payments.DataLocks.Application/Services/ApprenticeshipProcessor.cs
samuel-dev-dba: Name changed

PR #617, opened 2019-08-12
Add Metrics Payments stored procedure for period end Metrics report

jamie-coales:


ESFA-Pawan:
This was agreed to merge without testing as DC SP reviewed by Carl and there is no ticket.

PR #620, opened 2019-08-14
Release/release pv2 1346

ESFA-Pawan:
PV2_1346


Ron2404:
This can be merged into master . The caveat though is that it will need to be retested once the issues on the  MA side are resolved .  As of now data that is reflected in the API is being correctly picked up by PaymentsV2 and inserted in the Employer Levy accounts and Apprenticeship tables.

PR #628, opened 2019-08-18
Pv2 1287 submission succeeded event

davehowlett:


src/SFA.DAS.Payments.EarningEvents.Application/Handlers/JobContextMessageHandler.cs
```diff
@@ -71,6 +73,26 @@ public async Task<bool> HandleAsync(JobContextMessage message, CancellationToken
             logger.LogDebug($"Processing Earning Event Service event for Job Id : {message.JobId}");
             try
             {
+                if (message.Topics.Any())
```
MichaelYoung1981: Can this block be extracted out into a method describing what it's doing?

src/SFA.DAS.Payments.EarningEvents.Application/Handlers/JobContextMessageHandler.cs
```diff
@@ -71,6 +73,26 @@ public async Task<bool> HandleAsync(JobContextMessage message, CancellationToken
...
+                {
+                    var subscriptionMessage = message.Topics.SingleOrDefault(m => m.SubscriptionName == "GenerateFM36Payments");
+
+                    if (subscriptionMessage != null && subscriptionMessage.Tasks.Any())
+                    {
+                        if (subscriptionMessage.Tasks.Any(t => t.Tasks.Contains("JobSuccess")))
```
MichaelYoung1981: Can the hard coded strings be moved out?

Also, is it worth investigating a way to dispatch messages in a generic fashion with separate handlers?  There are 3 different message handlers in this class and there is potential for this to turn into a massive switch.

src/SFA.DAS.Payments.FundingSource.Application/Repositories/LevyFundingSourceRepository.cs
```diff
@@ -80,5 +80,15 @@ public async Task<List<long>> GetEmployerAccounts(CancellationToken cancellation
             accounts.AddRange(transferSenders);
             return accounts.Distinct().ToList();
         }
+        public async Task<long> GetLevyAccountId(long ukprn, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var levyAccounts = await dataContext.Payment.AsNoTracking()
+                .Where(payment => payment.Ukprn == ukprn)
+                .Where(payment => payment.AccountId.HasValue)
+                .Distinct()
+                .ToListAsync(cancellationToken);
```
MichaelYoung1981: Is the ToList necessary?  Looks like it's there to prevent deferred execution?

src/SFA.DAS.Payments.FundingSource.Application/Services/RequiredLevyAmountFundingSourceService.cs
```diff
@@ -19,6 +19,7 @@
 using SFA.DAS.Payments.Model.Core.Entities;
 using SFA.DAS.Payments.DataLocks.Messages.Events;
 using SFA.DAS.Payments.FundingSource.Model;
+using SFA.DAS.Payments.Model.Core;
 
 namespace SFA.DAS.Payments.FundingSource.Application.Services
 {
```
MichaelYoung1981: At what point is this class doing too much?  It's long and has a boat load of dependencies.

src/SFA.DAS.Payments.FundingSource.LevyFundedProxyService/Handlers/SubmissionSucceededEventHandler.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            if (message.Ukprn == 0)
+                throw new ArgumentException($"Ukprn cannot be 0. Job Id: {message.JobId}");
+
+            try
+            {
+                logger.LogDebug($"Getting AccountId for Ukprn: {message.Ukprn}.");
+                var accountId = await repository.GetLevyAccountId(message.Ukprn).ConfigureAwait(false);
+                var actorId = new ActorId(accountId);
+                var actor = proxyFactory.CreateActorProxy<ILevyFundedService>(new Uri("fabric:/SFA.DAS.Payments.FundingSource.ServiceFabric/LevyFundedServiceActorService"), actorId);
```
MichaelYoung1981: As this string is used in multiple places shouldn't it be a constant somewhere?

src/SFA.DAS.Payments.FundingSource.LevyFundedProxyService/Handlers/SubmissionSucceededEventHandler.cs
```diff
@@ -0,0 +1,69 @@
...
+using SFA.DAS.Payments.Application.Infrastructure.Logging;
+using SFA.DAS.Payments.FundingSource.Application.Interfaces;
+using SFA.DAS.Payments.FundingSource.LevyFundedService.Interfaces;
+using SFA.DAS.Payments.FundingSource.Messages.Internal.Commands;
+using SFA.DAS.Payments.Messages.Core.Events;
+using SFA.DAS.Payments.Model.Core.Factories;
+
+namespace SFA.DAS.Payments.FundingSource.LevyFundedProxyService.Handlers
+{
+    public class SubmissionSucceededEventHandler:IHandleMessages<SubmissionSucceededEvent>
```
MichaelYoung1981: The contents of this class appears almost identical to SubmissionFailedEventHandler, could the duplicate code be refactored out into a shared (internal) class?

src/SFA.DAS.Payments.ProviderPayments.ProviderPaymentsService/Handlers/SubmissionSucceededEventHandler.cs
```diff
@@ -0,0 +1,90 @@
...
+using AutoMapper;
+using NServiceBus;
+using SFA.DAS.Payments.Application.Infrastructure.Logging;
+using SFA.DAS.Payments.Messages.Core.Events;
+using SFA.DAS.Payments.Monitoring.Jobs.Client;
+using SFA.DAS.Payments.ProviderPayments.Application.Services;
+
+namespace SFA.DAS.Payments.ProviderPayments.ProviderPaymentsService.Handlers
+{
+    public class SubmissionSucceededEventHandler:IHandleMessages<SubmissionSucceededEvent>
```
MichaelYoung1981: Are there 2 copies of this class?

src/SFA.DAS.Payments.FundingSource.LevyFundedProxyService/Handlers/SubmissionSucceededEventHandler.cs
MichaelYoung1981: Should these handlers not have unit tests?

src/SFA.DAS.Payments.FundingSource.Application/Services/RequiredLevyAmountFundingSourceService.cs
```diff
@@ -191,6 +192,49 @@ public async Task<ReadOnlyCollection<FundingSourcePaymentEvent>> HandleMonthEnd(
...
+        {
+            var keys = await generateSortedPaymentKeys.GeyKeys().ConfigureAwait(false);
+
+            paymentLogger.LogDebug($"Processing {keys.Count} required payments, account {employerAccountId}, job id {jobId}");
+
+            foreach (var key in keys)
+            {
+                var requiredPaymentEvent = await requiredPaymentsCache.TryGet(key).ConfigureAwait(false);
+
+                if (requiredPaymentEvent.Value.CollectionPeriod == collectionPeriod &&
```
samuel-dev-dba: We need to also check the Academic Year

samuel-dev-dba: I don't think this will work, comparing objects that don't implement IEqualityComparer I will just compare the Year and Period

src/SFA.DAS.Payments.FundingSource.Application/Services/RequiredLevyAmountFundingSourceService.cs
```diff
@@ -191,6 +192,49 @@ public async Task<ReadOnlyCollection<FundingSourcePaymentEvent>> HandleMonthEnd(
...
+                    requiredPaymentEvent.Value.IlrSubmissionDateTime < submissionDate)
+                {
+                    await requiredPaymentsCache.Clear(key).ConfigureAwait(false);
+                }
+            }
+
+            paymentLogger.LogInfo("Finished removing previous submission payments.");
+        }
+
+        public async Task RemoveCurrentSubmission(long employerAccountId, long jobId, CollectionPeriod collectionPeriod,
```
samuel-dev-dba: This Look similar to RemovePreviousSubmissions I think we should have a base class and then each Implementation can override the IlrSubmissionDateTime logic.

src/SFA.DAS.Payments.FundingSource.LevyFundedProxyService/Handlers/SubmissionFailedEventHandler.cs
```diff
@@ -0,0 +1,69 @@
...
+                throw new ArgumentException($"Ukprn cannot be 0. Job Id: {message.JobId}");
+
+            try
+            {
+                logger.LogDebug($"Getting AccountId for Ukprn: {message.Ukprn}.");
+                var accountId = await repository.GetLevyAccountId(message.Ukprn).ConfigureAwait(false);
+                var actorId = new ActorId(accountId);
+                var actor = proxyFactory.CreateActorProxy<ILevyFundedService>(new Uri("fabric:/SFA.DAS.Payments.FundingSource.ServiceFabric/LevyFundedServiceActorService"), actorId);
+
+                var processCommand = new ProcessCurrentSubmissionDeletionCommand
```
samuel-dev-dba: I will not call this class a command as is not an NService Bus command this is just a POCO used in the actors. Actually, I don't think is needed can just pass in the event.

src/SFA.DAS.Payments.FundingSource.LevyFundedProxyService/Handlers/SubmissionSucceededEventHandler.cs
samuel-dev-dba: Yes, we should have a base handler class that they both override.

src/SFA.DAS.Payments.ProviderPayments.ProviderPaymentsService/Handlers/SubmissionSucceededEventHandler.cs
```diff
@@ -0,0 +1,90 @@
...
+            {
+                paymentLogger.LogError($"Error while handling SubmissionSucceededEvent event. Error: {ex.Message}, Job: {message.JobId}, UKPRN: {message.Ukprn}", ex);
+                throw;
+            }
+
+            paymentLogger.LogDebug($"Finished processing Submission Succeeded Event for Message Id : {context.MessageId}. ");
+        }
+    }
+
+    public class SubmissionFailedEventHandler : IHandleMessages<SubmissionFailedEvent>
```
samuel-dev-dba: Please create a base handler class

src/SFA.DAS.Payments.ProviderPayments.ProviderPaymentsService/Handlers/SubmissionSucceededEventHandler.cs
davehowlett: yes as it is being handled in both funding source and provider payments so there needs to be an implementation in each to handle the NSB message

src/SFA.DAS.Payments.FundingSource.Application/Services/RequiredLevyAmountFundingSourceService.cs
davehowlett: ok - changed

src/SFA.DAS.Payments.EarningEvents.Application/Handlers/JobContextMessageHandler.cs
davehowlett: yep. Created new method

src/SFA.DAS.Payments.EarningEvents.Application/Handlers/JobContextMessageHandler.cs
davehowlett: the actual message we get from DC is the same in all three cases - the only difference is within the Topics.Tasks list.

davehowlett: moving hardcoded strings out

src/SFA.DAS.Payments.FundingSource.Application/Repositories/LevyFundingSourceRepository.cs
davehowlett: I've just followed other methods in the class. Perhaps not necessary

src/SFA.DAS.Payments.FundingSource.Application/Services/RequiredLevyAmountFundingSourceService.cs
davehowlett: task to be raised to refactor after go live. PV2-1397

src/SFA.DAS.Payments.FundingSource.LevyFundedProxyService/Handlers/SubmissionSucceededEventHandler.cs
davehowlett: all our services are written like this with a string in the class consuming the actor service. I'll create a task on the backlog to deal with it. PV2-1398

src/SFA.DAS.Payments.FundingSource.LevyFundedProxyService/Handlers/SubmissionSucceededEventHandler.cs
davehowlett: none of our handlers have unit tests currently. We test them via the acceptance tests

davehowlett: I've created a base handler for both these handlers and renamed the POCO

src/SFA.DAS.Payments.FundingSource.Application/Services/RequiredLevyAmountFundingSourceService.cs
davehowlett: a base class for comparing == and < ? Not sure I understand?

src/SFA.DAS.Payments.FundingSource.LevyFundedProxyService/Handlers/SubmissionFailedEventHandler.cs
davehowlett: renamed

src/SFA.DAS.Payments.ProviderPayments.ProviderPaymentsService/Handlers/SubmissionSucceededEventHandler.cs
davehowlett: created

src/SFA.DAS.Payments.ProviderPayments.ProviderPaymentsService/Handlers/SubmissionSucceededEventHandler.cs
samuel-dev-dba: We can have a base Handler in the service fabric core project which can then be overridden by each service that needs to handle the message,  just to prevent duplicate code. I am flexible with that but makes the code cleaner.

src/SFA.DAS.Payments.FundingSource.Application/Services/RequiredLevyAmountFundingSourceService.cs
samuel-dev-dba: As the method is the same we should move to the base class and override the diffrent section of the code. Due to time constraint, I am happy to let this go for now.

src/SFA.DAS.Payments.ProviderPayments.ProviderPaymentsService/Handlers/SubmissionSucceededEventHandler.cs
samuel-dev-dba: still needs so changes

PR #640, opened 2019-08-23
Pv2 1412 take 2 fix for missing apprenticeship id in data lock status changed service

kuzinator:
Missing apprenticeship ID is changed to warning with the log of DataLockEvent to see why

### DIFF
--- a/src/SFA.DAS.Payments.FundingSource.Application/Repositories/LevyFundingSourceRepository.cs
+++ b/src/SFA.DAS.Payments.FundingSource.Application/Repositories/LevyFundingSourceRepository.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.EntityFrameworkCore;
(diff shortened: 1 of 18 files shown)


src/SFA.DAS.Payments.FundingSource.Application/Repositories/LevyFundingSourceRepository.cs
```diff
@@ -80,15 +79,17 @@ public async Task<List<long>> GetEmployerAccounts(CancellationToken cancellation
...
         {
-            var levyAccounts = await dataContext.Payment.AsNoTracking()
-                .Where(payment => payment.Ukprn == ukprn)
-                .Where(payment => payment.AccountId.HasValue)
-                .Distinct()
-                .ToListAsync(cancellationToken);
+            var lastPayment = await dataContext.Payment.AsNoTracking()
+                .Where(payment => payment.Ukprn == ukprn && payment.AccountId.HasValue)
+                .OrderByDescending(payment => payment.EventTime)
+                .Take(1)
```
samuel-dev-dba: There is no need for Take 1 if we are using first of Default.

src/SFA.DAS.Payments.FundingSource.LevyFundedProxyService/Handlers/SubmissionEventHandler.cs
```diff
@@ -42,12 +42,19 @@ public async Task Handle(T message, IMessageHandlerContext context)
             {
                 logger.LogDebug($"Getting AccountId for Ukprn: {message.Ukprn}.");
                 var accountId = await repository.GetLevyAccountId(message.Ukprn).ConfigureAwait(false);
-                var actorId = new ActorId(accountId);
+
+                if (!accountId.HasValue)
```
samuel-dev-dba: I think we also need to check the Transfer Account ID

PR #647, opened 2019-08-28
Pv2 1415 change monitoring service to use append only model for storing job message status updates

abroodie:
All comments will be fixed in ticket PV2-1426 unless they are issues that should stop the release.


DavidBSteele:
This is a duplicate PR.

PR #648, opened 2019-08-28
Pv2 1431 cancelled operation exceptions

abroodie:


src/SFA.DAS.Payments.DataLocks.Application.UnitTets/Services/DataLockEventProcessorTest.cs
```diff
@@ -378,61 +378,6 @@ public async Task TestDataLockStatusChangedToPassedWithNonPayablePeriods()
             statusChangedEvents[0].TransactionTypesAndPeriods.First().Value[0].Period.Should().Be(3);
         }
 
-        [Test] // PV2-1412
```
samuel-dev-dba: Why are we removing this unit test?

PR #649, opened 2019-08-29
Pv2 1399 remove month end check from provider events

kuzinator:
Tests are not in good shape, they fail a lot but succeed on individual runs. Subject to fixing in a spearate task.


src/SFA.DAS.Payments.DataLocks.Application/Services/DataLockEventProcessor.cs
```diff
@@ -239,13 +240,13 @@ private void AddTypeAndPeriodToEvent(DataLockStatusChanged statusChangedEvent, T
                 {
                     foreach (var period in onProgrammeEarning.Periods)
                     {
-                        if (IsEmpty(period))
```
samuel-dev-dba: Do we need this code changes for Month-end handling

ESFA-Pawan:
This is too old fix and recent month end for a small set didn't generate any exception in SDW. Agreed with Tony and Harpal to close this.

PR #650, opened 2019-08-29
Pv2-1400 PV2-1418 duplicate aims with incorrect contract type

CarlHA:


src/SFA.DAS.Payments.EarningEvents.Application/Mapping/MappingExtensions.cs
```diff
@@ -44,9 +44,14 @@ public static ContractType[] GetContractTypesForLearningDeliveries(this List<Lea
             for (byte i = 1; i <= earningPeriods; i++)
             {
                 var periodValues = periodisedTextValues.Select(p => p.GetPeriodTextValue(i)).ToArray();
-                var periodValue = GetContractType(periodValues[0]);
-
-                contractTypes[i - 1] = periodValue;
+                if (periodValues.All(p => p == "None"))
```
kuzinator: Please add protection against this happening in period 1

CarlHA: Updated

DavidBSteele:
This is not required as the story is still under analysis.

PR #653, opened 2019-09-02
Pv2 1359 duplicate payments generated after month end for act1 

abroodie:


src/SFA.DAS.Payments.FundingSource.Application/Repositories/LevyFundingSourceRepository.cs
```diff
@@ -80,16 +80,18 @@ public async Task<List<long>> GetEmployerAccounts(CancellationToken cancellation
...
-                .Where(payment => payment.Ukprn == ukprn && payment.AccountId.HasValue)
-                .OrderByDescending(payment => payment.EventTime)
-                .Take(1)
-                .FirstOrDefaultAsync(cancellationToken)
+            var accounts = await dataContext.Apprenticeship.AsNoTracking()
+                .Where(apprenticeship => apprenticeship.Ukprn == ukprn)
+                .Select (apprenticeship => new {
+                    apprenticeship.AccountId, 
+                    apprenticeship.TransferSendingEmployerAccountId
+                })
```
carlbm: IIRC, an employer can have some apprentices 'self-funded' and some by transfers. Does this structure allow for this case?

abroodie: Yes, the levy proxy service will use these 2 values to route the request to the correct levy actor.

DavidBSteele:
Dev meeting decided that this should be closed. It has already been released and merged to master. 1461 and 1462 have replaced it.

PR #654, opened 2019-09-02
pv2-949 Fix of Entity Framework issue relating to ReportingFundingAimLineType on Payment entity.

keenesolutions:


keenesolutions:
There is a mismatch between the DB schema and entity framework schema relating to the ReportingAimFundingLineType column. In the DB it's nullable but in EF it's not. This change will fix any tests where the PaymentModel's ReportingAimFundingLineType is null.

benagain:
This PR makes sense as a default setting for the tests.  I think it would also be a good idea to fix the the underlying schema mismatch.

keenesolutions:
The underlying schema mismatch is intentional. This is to prevent new records missing the ReportingAimFundingLineType  going forward.

PR #658, opened 2019-09-04
Pv2 1443 Test fixes for R01

keenesolutions:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/ApprovalsInterfaceSteps.cs
```diff
@@ -28,6 +28,19 @@ namespace SFA.DAS.Payments.AcceptanceTests.EndToEnd.Steps
     [Binding]
     public class ApprovalsInterfaceSteps : EndToEndStepsBase
     {
+
+        [BeforeStep()]
+        public void InitialiseNewTestDataContext()
+        {
+            TestDataContext = Scope.Resolve<TestPaymentsDataContext>();
```
benagain: Is this simply to force a reload from the db at every spec step?

keenesolutions: It's to prevent threading issues when tests are run in parrallel.

PR #672, opened 2019-09-10
PV2-1446_do_not_retry_on_contract_type_none

kuzinator:


src/SFA.DAS.Payments.EarningEvents.Application/Mapping/ApprenticeshipContractTypeEarningsEventFactory.cs
```diff
@@ -1,25 +1,21 @@
...
-        public const string ContractForServicesWithEmployer = "Contract for services with the employer";
-        public const string ContractForServicesWithSfa = "Contract for services with the ESFA";
-
         public ApprenticeshipContractTypeEarningsEvent Create(string contractType)
         {
             switch (contractType)
             {
-                case Act1:
-                case ContractForServicesWithEmployer:
+                case ContractTypeValidationRule.Act1:
```
carlbm: I don't think that the `ContractTypeValidationRule` is the correct place for the ACT1/2 values. Maybe move them to their own class in `Payments.Model.Core`?

src/SFA.DAS.Payments.EarningEvents.Domain.UnitTests/Validation/Learner/ContractTypeValidationRuleTest.cs
```diff
@@ -0,0 +1,52 @@
...
+                        PriceEpisodeIdentifier = "pe-1",
+                        PriceEpisodeValues = new PriceEpisodeValues
+                        {
+                            EpisodeStartDate = DateTime.Today.AddMonths(-2),
+                            PriceEpisodePlannedEndDate = DateTime.Today.AddDays(-30),
+                            PriceEpisodeActualEndDate = DateTime.Today.AddDays(-30),
+                            PriceEpisodeContractType = contractType
+                        }
+                    },
+                    new PriceEpisode
```
carlbm: Do you need this?

src/SFA.DAS.Payments.EarningEvents.Application/Mapping/MappingExtensions.cs
```diff
@@ -97,32 +96,5 @@ public static List<EarningPeriod> CreateIncentiveEarningPeriods(this PriceEpisod
             result.AddPeriodValue(values.Period12, 12, priceEpisodeIdentifier, 1);
             return result;
         }
-
-        public static PriceEpisode GetLatestOnProgPriceEpisode(this List<PriceEpisode> priceEpisodes)
```
carlbm: Were these methods just not used?

carlbm:
Something to discuss - if there is a validation failure doesn't that just throw an exception that is handled by NServiceBus through the normal retry mechanism?

benagain:
This approach is obsolete and has been replaced by #765

PR #675, opened 2019-09-11
PV2-1464_ DLOCK10 - Against a stopped Commitment which is due to course change or user inputting wrong details in

samuel-dev-dba:


src/SFA.DAS.Payments.DataLocks.Domain/Services/CourseValidation/EarningPeriodsValidationProcessor.cs
```diff
@@ -115,6 +124,42 @@ public class EarningPeriodsValidationProcessor : IEarningPeriodsValidationProces
...
+                }
+            }
+            else
+            {
+                apprenticeshipsToUseThisPeriod.AddRange(activeApprenticeships);
+            }
+
+            if (apprenticeshipsToUseThisPeriod.Count == 0)
+            {
+                apprenticeshipsToUseThisPeriod = apprenticeships;
```
benagain: I think this is saying: "If there are no apprenticeships that started in the requested period then return *all* apprenticeships."  Is this correct?

PR #678, opened 2019-09-11
pv2-1409 remove exception logging behaviour

benagain:
Improve the signal-to-noise ratio in the logs in order to see bone-fide errors with message processing.  Currently each and every message failure is logged as an Error, even if the message succeeds on a subsequent retry.

[NServiceBus' default behaviour](https://docs.particular.net/nservicebus/recoverability/#retry-logging) is to log all failures at level Warn, with only the final failure logged at level Error.  We have an NServiceBus behaviour that independently logs all failures as Error, and it seems to be this behaviour that is causing our Too Much Information problem.

As a first step we are going to remove this behaviour and revert to NServiceBus' default.  This drops the number of `Error` logs per actually-failed message from 24 to 1.

### DIFF
--- a/src/SFA.DAS.Payments.Application/Infrastructure/Ioc/Modules/MessagingModule.cs
+++ b/src/SFA.DAS.Payments.Application/Infrastructure/Ioc/Modules/MessagingModule.cs
@@ -5,6 +5,8 @@ using System.Web;
 using Autofac;
 using NServiceBus;
 using NServiceBus.Features;
+using NServiceBus.Logging;
+using SFA.DAS.Payments.Application.Infrastructure.Logging;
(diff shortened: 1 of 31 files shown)


benagain:
After some discussions yesterday I am going to look at keeping some of the extra information that has been removed from the `LogError` lines.

There are two ways this could be done:
1. `LogWarn` those lines instead.  This meets the general goal of only keeping 1 `Error` log per failed message.  However, `LogWarn` is not capable of logging the exception to the `exception` field of the log.

2. Throw a new `MessageProcessingFailed` exception whose message contains all the extra info, and let the normal logging take care of logging that as `info`/`warning` during retires and `error` only on the final failure.  This moves the original exception to an inner exception, so will only work if the those are logged properly.

PR #683, opened 2019-09-13
PV2-1471: Clear old jobs by ID

kuzinator:
clear job ID before sending test job


src/SFA.DAS.Payments.AcceptanceTests.Core/Data/TestPaymentsDataContext.cs
```diff
@@ -94,5 +94,10 @@ public void ClearApprenticeshipData(long apprenticeshipId)
 ";
             Database.ExecuteSqlCommand(deleteApprenticeshipData, apprenticeshipId);
         }
+
+        public void ClearJobId(long jobId)
+        {
+            Database.ExecuteSqlCommand($"delete from Payments2.Job where DCJobId = {jobId}");
```
samuel-dev-dba: I think we should also Delete from- Payments2].[JobEvent]

kuzinator: good point. This table is likely to go but for the sake of consistency I've added the line

PR #688, opened 2019-09-16
When Duplicate Apprenticeship Is Same Ukprn Return No DataLockErrors

samuel-dev-dba:


MichaelYoung1981:
I don't believe the logic implemented is correct as we'll still get incorrect DLOCK08 failures.  Can we catch up about it this morning?

PR #689, opened 2019-09-16
PV2-1481 - View of the Payments v2 Database for DC Data Match Report

benagain:
So that DC can create and view the Data Match report with minimal impact on the Payments v2 database performance.


MichaelYoung1981:
Does the view not also need to be included in the sqlproj file?

ESFA-Pawan:
Passed in DSI - DAS Payments V2-0.0.3906-R002
New view available. Need to raise minor bug for aim sequence number being 0 for acceptance tests. In shadow aim sequence number is not 0.

PR #698, opened 2019-09-19
Added earning adjustment statement table PV2-1171

MichaelYoung1981:


src/SFA.DAS.Payments.Database/Payments2/Tables/Payments2.EarningsAdjustmentStatement.sql
```diff
@@ -0,0 +1,13 @@
+﻿CREATE TABLE [Payments2].[EarningsAdjustmentStatement]
```
carlbm: Hi
Agreed table name was [ProviderAdjustmentPayments]

PR #704, opened 2019-09-19
Pv2-1475 Employer change of circumstances fix

keenesolutions:


keenesolutions:
I've updated this since my original commit so that employer Ids are only copied for act1. This is simplier and doesn't change any logic in the matchers.

PR #716, opened 2019-09-23
PV2-1498 - Speed up Monitoring of Large Providers

benagain:
Streamline the monitoring activity when processing a large provider through the Payments v2 monitoring service by creating specific nodes for the top 5 providers.


src/SFA.DAS.Payments.Monitoring.Jobs.Client/EarningsJobClient.cs
```diff
@@ -24,6 +24,7 @@ public class EarningsJobClient : IEarningsJobClient
         private readonly IMessageSession messageSession;
         private readonly IPaymentLogger logger;
         private readonly IConfigurationHelper config;
+        private readonly JobMonitorPartition partitionName = new JobMonitorPartition();
```
carlbm: Can you pass this in as a constructor argument instead? Would just make it more testable

src/SFA.DAS.Payments.Monitoring.Jobs.Client.UnitTests/PartitionNameTests.cs
```diff
@@ -0,0 +1,59 @@
...
+        [TestCase(72, 0, 12)]
+        [TestCase(73, 0, 13)]
+        [TestCase(74, 0, 14)]
+        [TestCase(75, 0, 15)]
+        [TestCase(76, 0, 16)]
+        [TestCase(77, 0, 17)]
+        [TestCase(78, 0, 18)]
+        [TestCase(79, 0, 19)]
+        [TestCase(80, 0, 0)]
+        [TestCase(1, 10003915, 10003915)]
```
carlbm: Can you use a different job id for the same UKPRN? Just to prove that this isn't a happy mistake

src/SFA.DAS.Payments.Monitoring.Jobs.Client/JobMonitorPartition.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Linq;
+
+namespace SFA.DAS.Payments.Monitoring.Jobs.Client
+{
+    public class JobMonitorPartition
+    {
+        public static long[] InvidivualNodes = new long[]
```
carlbm: Use a hashset instead? Really not 100% though :)

src/SFA.DAS.Payments.Monitoring.Jobs.Client/EarningsJobClient.cs
benagain: This is a good idea, when the tests are written :)

src/SFA.DAS.Payments.Monitoring.Jobs.Client/JobMonitorPartition.cs
benagain: I'm not sure it's worth it for a collection of 5 values?

src/SFA.DAS.Payments.Monitoring.Jobs.Client.UnitTests/PartitionNameTests.cs
benagain: Good idea, done 👍

PR #724, opened 2019-09-26
Pv2 1518 update the applications to use the nservice bus license from config

abroodie:


ESFA-Pawan:
DAS Payments V2-0.0.4154-R002 - released to prod. 
New build for this PR has been approved and only reverts changes which were done after the prod deployment.

PR #737, opened 2019-09-30
Only monitor messages required for monitoring

benagain:


src/SFA.DAS.Payments.Monitoring.Jobs.Application/JobProcessing/JobStatusService.cs
```diff
@@ -142,7 +142,8 @@ private async Task CompleteDataLocks(long jobId, List<CompletedMessage> complete
                 "EarningFailedDataLockMatching",
                 "ProcessLearnerCommand",
                 "Act1FunctionalSkillEarningsEvent",
-                "ApprenticeshipContractType1EarningEvent"
+                "ApprenticeshipContractType1EarningEvent",
```
samuel-dev-dba: This should be list of types rather than strings

benagain: Good idea.  Changed.

PR #742, opened 2019-10-02
PV2-1504_Incorrect_Mapping_Of_Funding_To_Reporting_Line_Types

samuel-dev-dba:


src/SFA.DAS.Payments.EarningEvents.Application.UnitTests/ApprenticeshipContractTypeEarningsEventBuilderTest.cs
```diff
@@ -0,0 +1,218 @@
...
+                                    Period7 = 0,
+                                    Period8 = 0,
+                                    Period9 = 0,
+                                    Period10 = 0,
+                                    Period11 = 0,
+                                    Period12 = 300,
+                                },
+                                new LearningDeliveryPeriodisedValues
+                                {
+                                    AttributeName = "PriceEpisodeLSFCash",
```
CarlHA: The correct attribute name is LearnSuppFundCash

samuel-dev-dba: changed

src/SFA.DAS.Payments.EarningEvents.Application.UnitTests/ApprenticeshipContractTypeEarningsEventBuilderTest.cs
```diff
@@ -0,0 +1,218 @@
...
+
+        [OneTimeSetUp]
+        public void InitialiseMapper()
+        {
+            mapper = new Mapper(new MapperConfiguration(cfg => cfg.AddProfile<EarningsEventProfile>()));
+        }
+
+
+        [Test]
+        public void BuildFundingLineTypeCorrectly()
```
benagain: Is there a more descriptive name that could help understand why this test is showing "correct" funding line types?

There is a lot going on in the setup and I don't know what is important to this test, what would cause the test to fail if it changed, etc.

PR #763, opened 2019-10-11
pv-1560 - Use correct variable in config; consolidate nuget packages

benagain:
When investigating some environment problems we noticed that the earning events cloud config used an incorrect config variable, and various versions of DC nuget were consumed by different projects.


src/SFA.DAS.Payments.AcceptanceTests.Components.Core/SFA.DAS.Payments.AcceptanceTests.Components.Core.csproj
```diff
@@ -90,6 +90,7 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
   <ItemGroup>
+    <None Include="app.config" />
```
samuel-dev-dba: do we need this

benagain: Nope, it's been reverted.

PR #771, opened 2019-10-16
Release/release pv2 1562 1566

ESFA-Pawan:
PV2-1562
PV2-1566


ESFA-Pawan:
This has been split as 1566 is passing but number of failures are quite high

PR #781, opened 2019-10-22
Pv2 1032 Refunds for SEM flag - change in SFA contribution percentage

keenesolutions:


src/SFA.DAS.Payments.RequiredPayments.Domain.UnitTests/Services/RequiredPaymentProcessorTests.cs
```diff
@@ -228,18 +228,23 @@ public void ThenTheHistoricPaymentsWithADifferentSfaContributionPercentageShould
                 paymentsDueService.Setup(x => x.CalculateRequiredPaymentAmount(It.IsAny<decimal>(), It.IsAny<List<Payment>>())).Returns(expectedAmount);
                 refundService
                     .Setup(svc => svc.GetRefund(
-                        It.Is<decimal>(amount => amount == 0),
+                        It.Is<decimal>(amount => amount == -5),
```
benagain: ❓ We are changing existing tests here, were they wrong?  

If possible, might it be better to leave the existing tests and create new unit tests to highlight the incorrect behaviour / fix?

src/SFA.DAS.Payments.RequiredPayments.Application/Processors/EarningEventProcessorBase.cs
```diff
@@ -109,7 +109,7 @@ public async Task<ReadOnlyCollection<PeriodisedRequiredPaymentEvent>> HandleEarn
                         }
                     }
 
-                    if (requiredPayments.Sum(x => x.Amount) == 0)
+                    if (requiredPayments.Sum(x => x.Amount) == 0 && !requiredPayments.Any(rp=>rp.Amount > 0))
```
benagain: 🤔 This takes a little bit of thinking about to understand what's going on.

"If the sum of all payments is zero and no single payment is positive" ?

It's not clear to me why we would be checking this, I think it would be worth creating a descriptively named function to perform the check, or at least a named variable.  This would aid understanding of the purpose for skipping the rest of this period.

src/SFA.DAS.Payments.RequiredPayments.Domain/Services/RequiredPaymentProcessor.cs
```diff
@@ -62,7 +62,7 @@ private List<RequiredPayment> RefundPaymentsWithDifferentSfaContribution(decimal
             return paymentHistory
                 .Where(payment => payment.SfaContributionPercentage != currentSfaContributionPercentage)
                 .GroupBy(payment => payment.SfaContributionPercentage)
-                .SelectMany(group => refundService.GetRefund(0, group.ToList()))
+                .SelectMany(group => refundService.GetRefund(group.Sum(g=>g.Amount)  * -1, group.ToList()))
```
samuel-dev-dba: What will happen if Sum of all historic payments is < 0

src/SFA.DAS.Payments.RequiredPayments.Application/Processors/EarningEventProcessorBase.cs
keenesolutions: I'll change this to requiredPayments.All(rp=>rp == 0). I think this is clearer

src/SFA.DAS.Payments.RequiredPayments.Domain/Services/RequiredPaymentProcessor.cs
keenesolutions: I this sum is less than 0 then the negative number * -1 will result in a positive value. The Refund service does not create payments for 0 or positive values so nothing will be generated.

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/RefundsSteps.cs
```diff
@@ -50,8 +50,7 @@ public void GivenThePreviouslySubmittedTheFollowingLearnerDetails(string provide
 
         [Given(@"the following earnings had been generated for the learner")]
         public void GivenTheFollowingEarningsHadBeenGeneratedForTheLearner(Table table)
-        {
-            PreviousEarnings = CreateEarnings(table, TestSession.Provider.Ukprn);
+        { PreviousEarnings = CreateEarnings(table, TestSession.Provider.Ukprn);
```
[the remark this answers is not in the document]
keenesolutions: yep. I'll revert this.

src/SFA.DAS.Payments.RequiredPayments.Application/Processors/EarningEventProcessorBase.cs
benagain: 👍  I like that much better :)

benagain (dismissed):

src/SFA.DAS.Payments.RequiredPayments.Domain.UnitTests/Services/RequiredPaymentProcessorTests.cs
keenesolutions: The test were incorrect as they were setting up mock to use 0 amounts. The real service requires minus amount to pay refunds.

samuel-dev-dba (dismissed):

PR #790, opened 2019-10-30
Pv2 1583 Allow payments for Non Levy learners funded by a Transfer so that a DLOCK_11 is not triggered

samuel-dev-dba:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Helpers/ApprenticeshipHelper.cs
```diff
@@ -89,6 +89,10 @@ public static ApprenticeshipModel CreateApprenticeshipModel(Apprenticeship appre
                            default(DateTime?) :
                           apprenticeshipSpec.StopEffectiveFrom.ToDate(),
                 IsLevyPayer = employer.IsLevyPayer,
+                ApprenticeshipEmployerType = string.IsNullOrWhiteSpace(apprenticeshipSpec.EmployerType)
+                    ? ApprenticeshipEmployerType.Levy
+                    : GetNonNullableApprenticeshipEmployerTypeOnApproval(apprenticeshipSpec.EmployerType)
```
benagain: ❓ This seems a little strange...

* If `IsLevyPayer` does not contain any non-whitespace chars, then we use `ApprenticeshipEmployerType.Levy`.
* If `IsLevyPayer` is the exact text `Levy` then we use `ApprenticeshipEmployerType.Levy`.
* If `IsLevyPayer` contains any non-whitespace chars, *but* is any other text then we use `ApprenticeshipEmployerType.NonLevy`.

To me it seems that we should maybe say "match `NonLevy` exactly, or use `Levy` otherwise"?

And, why is this place different to anywhere else that uses `GetNonNullableApprenticeshipEmployerTypeOnApproval`?  Should `GetNonNullableApprenticeshipEmployerTypeOnApproval` have that behaviour?

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/ApprovalsInterfaceSteps.cs
```diff
@@ -754,5 +732,18 @@ private static bool MatchPriceEpisodes(List<ApprovalsApprenticeship.PriceEpisode
...
+        public static ApprenticeshipEmployerType? GetApprenticeshipEmployerTypeOnApproval(string employerType)
+        {
+            switch (employerType)
+            {
+                case "Levy":
+                    return ApprenticeshipEmployerType.Levy;
+                case "Non-Levy":
+                    return ApprenticeshipEmployerType.NonLevy;
+                default:
+                    return default(ApprenticeshipEmployerType?);
```
benagain: ⛏ Tiny style suggestion: we don't need to specify the type of the default, c# knows that already.  You can just say 

````c#
    return default;

File: src/SFA.DAS.Payments.DataLocks.Domain.UnitTests/Services/CourseValidation/ApprenticeshipLevyPayerValidatorTest.cs [hunk 3]
```diff
@@ -77,6 +77,39 @@ public void GivenApprenticeshipEmployerIsNotALevyPayerReturnDataLock_11()
             result.ApprenticeshipPriceEpisodes.Should().BeEmpty();
         }
 
-     
+        [Test]
+        public void WhenApprenticeshipEmployerNotIsALevyPayerButTransferReceiverReturnNoDataLockErrors()
```
benagain: 🔧 That's not grammatically correct English, maybe reword it?

❌  When Apprenticeship Employer Not Is A Levy Payer But Transfer Receiver Return No Data Lock Errors
✔️  When Apprenticeship Employer Is Not A Levy Payer But Transfer Receiver Then Return No Data Lock Errors

src/SFA.DAS.Payments.DataLocks.Domain.UnitTests/Services/CourseValidation/ApprenticeshipLevyPayerValidatorTest.cs
```diff
@@ -77,6 +77,39 @@ public void GivenApprenticeshipEmployerIsNotALevyPayerReturnDataLock_11()
...
+                    }
+                }
+            };
+
+            var validator = new ApprenticeshipLevyPayerValidator();
+            var result = validator.Validate(validation);
+            result.DataLockErrorCode.Should().BeNull();
+            result.ApprenticeshipPriceEpisodes.Should().HaveCount(1);
+            result.ApprenticeshipPriceEpisodes[0].Id.Should().Be(100);
+        }
```
benagain: :question: The test name says there should be no data lock errors, but doesn't say anything about price episodes.  Are those last two checks relevant to this test?

samuel-dev-dba: Yes, as part of the definition of no DataLock is that you have at least one valid price episode

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/ApprovalsInterfaceSteps.cs
samuel-dev-dba: True but that will fail to build as we have not made the changes in our build definition to use the latest C# features.

src/SFA.DAS.Payments.DataLocks.Domain.UnitTests/Services/CourseValidation/ApprenticeshipLevyPayerValidatorTest.cs
samuel-dev-dba: Good spot, now changed

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Helpers/ApprenticeshipHelper.cs
samuel-dev-dba: Changed

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/ApprovalsInterfaceSteps.cs
benagain: We should update that at some point! 😄

PR #791, opened 2019-10-30
PV2-1600 - Apply refund to correct reporting funding line 

benagain:


src/SFA.DAS.Payments.RequiredPayments.Application/Mapping/RequiredPaymentsProfile.cs
```diff
@@ -206,8 +209,6 @@ public RequiredPaymentsProfile()
                 .Ignore(x => x.InstalmentAmount)
                 .Ignore(x => x.NumberOfInstalments)
                 .Ignore(x => x.LearningStartDate)
-                .Ignore(x => x.ApprenticeshipId)
-                .Ignore(x => x.ApprenticeshipPriceEpisodeId)
                 .Ignore(x => x.NumberOfInstalments)
                 .Ignore(x => x.ApprenticeshipEmployerType)
```
carlbm: This will need to go

carlbm: I actually have no idea what would happen here...

src/SFA.DAS.Payments.RequiredPayments.Application/Processors/EarningEventProcessorBase.cs
```diff
@@ -125,6 +123,8 @@ public async Task<ReadOnlyCollection<PeriodisedRequiredPaymentEvent>> HandleEarn
                         mapper.Map(earningEvent, requiredPaymentEvent);
                         mapper.Map(requiredPayment, requiredPaymentEvent);
 
+                        requiredPaymentEvent.ApprenticeshipEmployerType = requiredPayment.ApprenticeshipEmployerType;
```
carlbm: I think that we should do a check. If the required payment has an `Amount < 0` then apply this line.
If positive then the properties are set in datalocks - but this method is used for all payments, including ACT2 etc. Leaving this here risks an ACT2 learner who used to be ACT1 etc being incorrectly funded.

carlbm: Also, we should include `ApprenticeshipId` and `ApprenticeshipPriceEpisodeId` in the copy

src/SFA.DAS.Payments.RequiredPayments.Domain/Services/RefundService.cs
```diff
@@ -43,8 +47,10 @@ public List<RequiredPayment> GetRefund(decimal amount, List<Payment> paymentHist
                         SfaContributionPercentage = group.Key.SfaContributionPercentage,
                         PriceEpisodeIdentifier = group.Key.PriceEpisodeIdentifier,
                         AccountId = group.Key.AccountId,
-                        TransferSenderAccountId = group.Key.TransferSenderAccountId
-
+                        TransferSenderAccountId = group.Key.TransferSenderAccountId,
+                        ApprenticeshipEmployerType = group.Key.ApprenticeshipEmployerType,
```
samuel-dev-dba: I had a look at the GetPaymentHistory Method and we don't seem to be loading (ApprenticeshipEmployerType) from the Database.  I think we need additional code changes to load this field from DB.

src/SFA.DAS.Payments.RequiredPayments.Application/Mapping/RequiredPaymentsProfile.cs
benagain: The `Ignore` overrides the `Map`

src/SFA.DAS.Payments.RequiredPayments.Application/Processors/EarningEventProcessorBase.cs
benagain: Added `AddRefundCommitmentDetails`

src/SFA.DAS.Payments.RequiredPayments.Domain/Services/RefundService.cs
benagain: You're right (of course 😄) - I've made `GetPaymentHistory` fill in the commitment properties too now.

PR #797, opened 2019-11-01
Fix Contract Type for Refund

samuel-dev-dba:


src/SFA.DAS.Payments.Audit.Application.UnitTests/Mapping/RequiredPayments/CalculatedRequiredCoInvestedAmountMappingTests.cs
```diff
@@ -0,0 +1,30 @@
...
+            return new CalculatedRequiredCoInvestedAmount
+            {
+                ContractType = ContractType.Act1,
+                ApprenticeshipId = 400L,
+                ApprenticeshipPriceEpisodeId = 800L
+            };
+        }
+
+
+        [Test]
```
benagain: Might it be worth testing all values on `ContractType` are mapped OK?

I'm thinking that because the original bug was a hard-coded value in the mapping, it would be worth writing the test such that different contract types are mapped correctly, rather than just a single type.

samuel-dev-dba: done

PR #805, opened 2019-11-11
PV2-1626 incorrect refunds on learner ref change

CarlHA:
Added an additional test for this scenario.  The fix has been made on #804

### DIFF
--- a/src/SFA.DAS.Payments.RequiredPayments.Application.UnitTests/Application/Processors/RefundRemovedLearningAimProcessorTests.cs
+++ b/src/SFA.DAS.Payments.RequiredPayments.Application.UnitTests/Application/Processors/RefundRemovedLearningAimProcessorTests.cs
@@ -291,9 +291,9 @@ namespace SFA.DAS.Payments.RequiredPayments.Application.UnitTests.Application.Pr
             // assert
             using (new AssertionScope())
             {
-                refunds.Count.Should().Be(2);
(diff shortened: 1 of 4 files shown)


src/SFA.DAS.Payments.RequiredPayments.Application.UnitTests/Application/Processors/RefundRemovedLearningAimProcessorTests.cs
```diff
@@ -250,5 +252,91 @@ public async Task Refunds_All_Required_Payments()
...
+                                .RefundLearningAim(identifiedLearner,
+                                                   mockPaymentHistoryCache.Object,
+                                                   CancellationToken.None)
+                                .ConfigureAwait(false);
+
+            // assert
+            using (new AssertionScope())
+            {
+                refunds.Count.Should().Be(2);
+                refunds.Cast<CalculatedRequiredLevyAmount>().Count(x => x.OnProgrammeEarningType == OnProgrammeEarningType.Learning).Should().Be(1);
```
carlbm: Can you use `Should().Contain(...` in this situation?

CarlHA: Updated syntax

PR #813, opened 2019-11-13
Pv2 1581 unable to accept price change in apprenticeship service

samuel-dev-dba:


src/SFA.DAS.Payments.DataLocks.Application.UnitTets/Services/DataLockEventProcessorTest.cs
```diff
@@ -378,6 +378,220 @@ public async Task TestDataLockStatusChangedToPassedWithNonPayablePeriods()
             statusChangedEvents[0].TransactionTypesAndPeriods.First().Value[0].Period.Should().Be(3);
         }
 
+        [Test]
+        public async Task TestDataLockStatusChangedToPassedWithPayablePeriods()
+        {
```
benagain: 🌱 I find tests with nearly 200 lines of setup code really hard understand what they are testing and why.  

I think we need to think about improving our setup of unit tests, by using `Builder` objects, removing mocks, and generally simplifying everything.

For example: in [EarningsEventBuilderTests](https://github.com/SkillsFundingAgency/das-payments-V2/blob/master/src/SFA.DAS.Payments.EarningEvents.Domain.UnitTests/EarningsEventBuilderTests.cs#L34) a standard `ProcessLearnerCommand`, with all the required basic data, is used as a starting point and then the test-specific changes are made.  This makes it incredibly easy to see why that particular test is different to other tests around `ProcessLearnerCommand`s.

That doesn't necessarily have to be the way forward, but something has to be done to improve our very large test setup sections.  It isn't this particular PR that's the problem, it's the unit tests more generally.

src/SFA.DAS.Payments.DataLocks.Application/Services/DataLockEventProcessor.cs
```diff
@@ -130,47 +130,63 @@ public async Task<List<DataLockStatusChanged>> ProcessDataLockFailure(DataLockEv
...
-                    if (newPassesGroupedByTypeAndPeriod.TryGetValue(
-                        (oldFailure.TransactionType, oldFailure.DeliveryPeriod), out var newPass))
+                    foreach (var (transactionType, period) in newPassesGroupedByTypeAndPeriod.Keys)
                     {
-                        AddTypeAndPeriodToEvent(changedToPassed, oldFailure.TransactionType, newPass, payableEarningEvent);
-                        failuresToDelete.Add(oldFailure.Id);
+                        if (newPassesGroupedByTypeAndPeriod.TryGetValue((transactionType, period), out var newPass))
+                        {
+                            AddTypeAndPeriodToEvent(changedToPassed, transactionType, newPass, payableEarningEvent);
+                        }
```
benagain: ❓ If there are some `oldFailures`; and any `newPassesGroupedByTypeAndPeriod` that were not `oldFailures`; then I think they will not have `AddTypeAndPeriodToEvent` called for them.  Is that correct behaviour, or would we want *all* `newPasses...` to be added whether or not they were previously failed?

i.e. `OldFailures = [a,b,c]` and `newPasses = [a,b,d]`
We would only call `AddTypeAndPeriodToEvent` for `a` and `b`, not `d`

Given that we are explicitly adding *all* `newPasses` if there were no `oldFailures`, should we not add them all when there are *some* `oldFailures`?

samuel-dev-dba: Good point now changes

src/SFA.DAS.Payments.DataLocks.Application/Services/DataLockEventProcessor.cs
```diff
@@ -130,47 +130,60 @@ public async Task<List<DataLockStatusChanged>> ProcessDataLockFailure(DataLockEv
...
-                    payableEarningEvent.LearningAim.ProgrammeType,
-                    payableEarningEvent.LearningAim.StandardCode,
+                    allTransactionTypes,
                     payableEarningEvent.LearningAim.Reference,
                     payableEarningEvent.CollectionYear
                 ).ConfigureAwait(false);
 
-                foreach (var oldFailure in oldFailures)
+
+                if (oldFailures.Any())
```
benagain: This check isn't really necessary as `foreach` will simply do nothing when the collection is empty.

samuel-dev-dba:
Closing in favour of PR - 826

PR #816, opened 2019-11-14
PV2-1634 no learning start date for removed learner/aims

carlbm:
…ner/aims


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndStepsBase.cs
```diff
@@ -328,6 +328,12 @@ protected List<PaymentModel> CreatePayments(ProviderPayment providerPayment, Lis
                 ?.FirstOrDefault(x => x.LearnerId == learnerTraining.LearnerId)
                 ?.PriceEpisodeId.ParseAsNullableLong();
 
+            DateTime? startDate = null;
+            if (!string.IsNullOrWhiteSpace(learnerTraining.StartDate))
+            {
+                startDate = learnerTraining.StartDate.ToNullableDate();
```
benagain: `ToNullableDate()` [already performs ](https://github.com/SkillsFundingAgency/das-payments-V2/blob/master/src/SFA.DAS.Payments.Tests.Core/SpecDateExtensions.cs#L28)the null / whitespace check, it seems unnecessary to repeat it here 😉

src/SFA.DAS.Payments.RequiredPayments.Domain.UnitTests/Services/NegativeEarningsServiceTests.cs
```diff
@@ -15,14 +15,12 @@ namespace SFA.DAS.Payments.RequiredPayments.Domain.UnitTests.Services
     public class NegativeEarningsServiceTests
     {
         private NegativeEarningsService sut;
-        private Mock<IRefundService> refundService;
```
benagain: 💯

src/SFA.DAS.Payments.RequiredPayments.Domain.UnitTests/Services/RefundRemovedLearningAimServiceTests.cs
```diff
@@ -12,72 +11,44 @@ namespace SFA.DAS.Payments.RequiredPayments.Domain.UnitTests.Services
...
     {
-        private AutoMock mocker;
         private List<Payment> history;
+        private RefundRemovedLearningAimService sut;
 
         [SetUp]
         public void SetUp()
         {
-            mocker = AutoMock.GetLoose();
+            sut = new RefundRemovedLearningAimService(new RefundService(), new PaymentDueProcessor());
```
benagain: I like these changes!

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndStepsBase.cs
carlbm: I learn something new that I don't like about `string` every day 👍 
Updated

src/SFA.DAS.Payments.RequiredPayments.Domain.UnitTests/Services/RefundRemovedLearningAimServiceTests.cs
MichaelYoung1981: I don't...

I'm all for not mocking the world but if we're not mocking them from a unit test that would suggest to me that they're internals and shouldn't be passed to the constructor?

src/SFA.DAS.Payments.RequiredPayments.Application/Mapping/RequiredPaymentsProfile.cs
```diff
@@ -29,7 +29,9 @@ public RequiredPaymentsProfile()
                 .ForMember(payment => payment.InstalmentAmount, opt => opt.MapFrom(episode => episode.InstalmentAmount))
                 .ForMember(payment => payment.NumberOfInstalments, opt => opt.MapFrom(episode => episode.NumberOfInstalments))
                 .ForMember(payment => payment.AccountId, opt => opt.MapFrom(episode => episode.AccountId))
-                .ForMember(payment => payment.TransferSenderAccountId, opt => opt.MapFrom(episode => episode.TransferSenderAccountId))
+                .ForMember(payment => payment.TransferSenderAccountId, opt => opt.MapFrom(episode => episode. TransferSenderAccountId))
```
MichaelYoung1981: If we're using AutoMapper but explicitly configuring every single property (even when the names match) there isn't anything Auto about it and we're just adding complexity.

src/SFA.DAS.Payments.RequiredPayments.Domain/Services/RefundRemovedLearningAimService.cs
```diff
@@ -10,10 +9,10 @@ public class RefundRemovedLearningAimService: IRefundRemovedLearningAimService
         private readonly IRefundService refundService;
         private readonly IPaymentDueProcessor paymentDueProcessor;
 
-        public RefundRemovedLearningAimService(IRefundService refundService, IPaymentDueProcessor paymentDueProcessor)
+        public RefundRemovedLearningAimService()
```
samuel-dev-dba: There is no need for this change DI will inject a new instance for us, so what is the point of creating a new instance here.

MichaelYoung1981: Because the "dependencies" shouldn't be public classes.  They've remained public because they're used in other places at the minute.

carlbm:
Decided that this is not required

PR #818, opened 2019-11-18
Fix of EarningEventId mapping in RequiredPaymentEvent table  for math and English.

keenesolutions:


src/SFA.DAS.Payments.RequiredPayments.Application/Mapping/RequiredPaymentsProfile.cs
```diff
@@ -118,6 +119,7 @@ public RequiredPaymentsProfile()
                 ;
 
             CreateMap<PayableFunctionalSkillEarningEvent, CalculatedRequiredIncentiveAmount>()
+                .ForMember(x => x.EarningEventId, opt => opt.MapFrom(src=>src.EarningEventId))
```
benagain: Looks fine, but I don't understand why it's needed.  AutoMapper should map identically named properties by default.

keenesolutions: Your correct, it should do but this fixed the issue. I think there is a conflicting mapping elsewhere tha is inplicitly ignoring the EarningEventId which is why if wasn't being honoured. The only fix was to do an explicit include. See lines 77/ 79.

ESFA-Pawan:
No change in the PR as code was also included in 1571 which is already in master. Merging to keep the history

PR #826, opened 2019-11-22
Pv2 1581 detect changes to price episodes

samuel-dev-dba:


src/SFA.DAS.Payments.DataLocks.Application/Services/PriceEpisodesReceivedService.cs
```diff
@@ -0,0 +1,152 @@
...
+            ICurrentPriceEpisodeForJobStore store,
+            IReceivedDataLockEventStore receivedEventStore,
+            PriceEpisodeStatusChangeBuilder statusChangeBuilder)
+        {
+            currentPriceEpisodesStore = store;
+            this.receivedEventStore = receivedEventStore;
+            this.statusChangeBuilder = statusChangeBuilder;
+        }
+
+        public async Task<List<PriceEpisodeStatusChange>> JobSucceeded(long jobId, long ukprn)
```
MichaelYoung1981: My only issue is that this method (and indeed the class name) doesn't help an unfamiliar reader (i.e. me) understand what the intent of the class and method is.

benagain: True.  The class is supposed to inform approvals of changes to price episodes.  Maybe `PriceEpisodeChangeNotifier`?

samuel-dev-dba: The method name is now changed to  GetPriceEpisodeChanges

PR #827, opened 2019-11-22
PV2-1663 Resolved issue of unexpected clawbacks on incentives

CarlHA:
Created feature file to replicate issue
Reduced the number of incentive periods inline with OnProg
Ensure LSF correctly include when in M&E aim

This PR also addresses issues in PV2-802/803 and PV2-805/806 which are all related to Learning Support


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndStepsBase.cs
```diff
@@ -292,8 +292,16 @@ protected List<PaymentModel> CreatePayments(ProviderPayment providerPayment, Lis
             var otherTraining = learnerTraining.FirstOrDefault(t => t.AimReference != "ZPROG001" && (string.IsNullOrWhiteSpace(t.LearnerId) || t.LearnerId.Equals(providerPayment.LearnerId)));
             var list = new List<PaymentModel>();
             if (providerPayment.SfaFullyFundedPayments > 0)
-                list.Add(CreatePaymentModel(providerPayment, otherTraining ?? onProgTraining, jobId, submissionTime, 1M,
-                    providerPayment.SfaFullyFundedPayments, FundingSourceType.FullyFundedSfa, ukprn, providerPayment.AccountId, providerPayment.SendingAccountId));
+                if (providerPayment.TransactionType == TransactionType.LearningSupport)
```
MichaelYoung1981: The complexity of this class makes me sad.  Not necessarily one for this PR but something to look at in the future.

src/SFA.DAS.Payments.EarningEvents.Application.UnitTests/FunctionalSkillEarningEventBuilderTest.cs
```diff
@@ -804,5 +804,271 @@ public void MapFundingLineTypeCorrectly()
...
+            var events = builder.Build(processLearnerCommand);
+
+            // assert
+            events.Should().NotBeNull();
+            events.Should().HaveCount(1);
+            events.Single().Earnings.Should().HaveCount(2);
+            events.Single().Earnings.Single(x => x.Type == FunctionalSkillType.LearningSupport).Periods.Should().HaveCount(12);
+        }
+
+        private static ProcessLearnerCommand CreateLearnerSubmissionWithLearningSupport()
```
MichaelYoung1981: Does the test rely on all of this specific detail or could we use some kind of base method (or an actual Builder) to hydrate the base object and just override the specifics needed for the test?

src/SFA.DAS.Payments.EarningEvents.Application/Mapping/FunctionalSkillsEarningValueResolver.cs
```diff
@@ -27,6 +27,7 @@ public ReadOnlyCollection<FunctionalSkillEarning> Resolve(IntermediateLearningAi
                 .Where(periodisedValues => TypeMap.ContainsKey(periodisedValues.AttributeName))
                 .GroupBy(v => v.AttributeName)
                 .Select(CreateEarning)
+                .Where(x => x.Periods.Any(p => p.Amount != 0))
```
MichaelYoung1981: Do we have a negative test for this? Or an explicit positive.

src/SFA.DAS.Payments.DataLocks.Domain/Services/CourseValidation/EarningPeriodsValidationProcessor.cs
```diff
@@ -167,7 +167,7 @@ private List<ApprenticeshipModel> GetApprenticeshipsToUseThisPeriod(long ukprn,
         private bool IsFunctionalSkillTransactionType(TransactionType transactionType)
         {
             var functionalSkillTransactionTypes = new List<TransactionType>
-                {TransactionType.OnProgrammeMathsAndEnglish, TransactionType.BalancingMathsAndEnglish};
+                {TransactionType.OnProgrammeMathsAndEnglish, TransactionType.BalancingMathsAndEnglish, TransactionType.LearningSupport};
```
samuel-dev-dba: This is really dangerous as OnProg Incentives also has Learning Support so we could think OnProg Incentive is a functional skill.

src/SFA.DAS.Payments.EarningEvents.Application/Mapping/IncentiveEarningValueResolver.cs
```diff
@@ -46,18 +50,33 @@ private static string IncentiveEarningTypeToFM36AttributeName(IncentiveEarningTy
 
         private IncentiveEarning CreateIncentiveEarning(List<PriceEpisode> priceEpisodes, IncentiveEarningType incentiveType)
         {
+            var attributeName = IncentiveEarningTypeToFM36AttributeName(incentiveType);
```
samuel-dev-dba: I think as this method is fundamental to incentives payments, can we move this logic into a separate class IncentiveEarningBuilder, have a test that describe the behaviour and call it in this method. 

Makes it easier to test.

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndStepsBase.cs
CarlHA: Definitely not one for this PR

src/SFA.DAS.Payments.EarningEvents.Application.UnitTests/FunctionalSkillEarningEventBuilderTest.cs
CarlHA: A FM36Learner is used in quite a few of the Unit Tests and does require some specific logic/rules.  I agree it should be in a builder, but it should be done properly as a number of unit test have had incorrect data in the FM36Learner and given incorrect test results.  This really should be raised as a new ticket and not as part of this PR.

MichaelYoung1981: I disagree, the code was added as part of this PR so it becomes part of this work surely?

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndStepsBase.cs
MichaelYoung1981: I'll create a tech debt item. 👍

src/SFA.DAS.Payments.EarningEvents.Application/Mapping/FunctionalSkillsEarningValueResolver.cs
CarlHA: There are a number of tests covering the positive case ie periods where the value != 0.  A negative test has been added.

src/SFA.DAS.Payments.EarningEvents.Application/Mapping/IncentiveEarningValueResolver.cs
CarlHA: This method is equivalent to the one in OnProgrammeEarningValueResolver and in both cases it is the fundamental purposes of class and it makes no sense to move it in to its own class. Both resolvers have 36 tests covering this area of the code.

src/SFA.DAS.Payments.EarningEvents.Application.UnitTests/FunctionalSkillEarningEventBuilderTest.cs
CarlHA: Its a question of priorities - this is not a trivial piece of work to replace all the instances of ProcessLearnerCommand/FM36Learner with a valid builder and unless it is done correctly we will perpetuate incorrect implementations because the data has been set up incorrectly.

src/SFA.DAS.Payments.EarningEvents.Application/Mapping/IncentiveEarningValueResolver.cs
samuel-dev-dba: You are right there is no need to move the code out but we can Have a test for - IncentiveEarningValueResolver class separately that will allow us to test the Resolve Method

src/SFA.DAS.Payments.DataLocks.Domain/Services/CourseValidation/EarningPeriodsValidationProcessor.cs
CarlHA: Renamed method to IsOnProgrammeIncentiveTransactionType

src/SFA.DAS.Payments.EarningEvents.Application.UnitTests/FunctionalSkillEarningEventBuilderTest.cs
MichaelYoung1981: I agree but how do we stop this complexity becoming further widespread?  We have to draw a line somewhere (possibly bad luck if we decide it's here! 😁).  To keep the size of the work to a minimum we could create a builder which is only used by this test class (but still explicitly override defaults required for this test) and, rather than changing all usages of ProcessLearnerCommand/FM36Learner, communicate to the team that the builder needs to be used going forward and in future code reviews enforce migration?

src/SFA.DAS.Payments.DataLocks.Domain/Services/CourseValidation/EarningPeriodsValidationProcessor.cs
CarlHA: This method has been removed following the latest refactoring

src/SFA.DAS.Payments.DataLocks.Domain/Services/CourseValidation/EarningPeriodsValidationProcessor.cs
```diff
@@ -81,7 +80,7 @@ public class EarningPeriodsValidationProcessor : IEarningPeriodsValidationProces
                     {
                         EarningPeriod = period,
                         Apprenticeship = apprenticeship,
-                        PriceEpisode = IsFunctionalSkillTransactionType(transactionType) 
+                        PriceEpisode = priceEpisodes == null
```
samuel-dev-dba: we can replace this code with =>  priceEpisodes?.SingleOrDefault

src/SFA.DAS.Payments.EarningEvents.Application/Mapping/IncentiveEarningValueResolver.cs
samuel-dev-dba: Reluctantly letting this go as there has been an extensive manual test on this code and we don't want to go back and re-test due to code changes.

src/SFA.DAS.Payments.DataLocks.Domain/Services/CourseValidation/EarningPeriodsValidationProcessor.cs
samuel-dev-dba: Reluctantly letting this go as there has been an extensive manual test on this code and we don't want to go back and re-test due to code changes.

PR #831, opened 2019-11-28
Pv2-1595 enum cast error on removed aim

keenesolutions:


src/SFA.DAS.Payments.RequiredPayments.Application/Processors/RefundRemovedLearningAimProcessor.cs
```diff
@@ -65,6 +65,13 @@ private IList<PeriodisedRequiredPaymentEvent> CreateRefundPayments(IdentifiedRem
                         throw new InvalidOperationException($"Cannot find historic payment with price episode identifier: {refund.payment.PriceEpisodeIdentifier} for period {refund.deliveryPeriod}.");
                     
                     var requiredPaymentEvent = requiredPaymentEventFactory.Create(refund.payment.EarningType, transactionType);
+                    if (requiredPaymentEvent == null)
+                    {
+                        // This shouldn't now happen as the transaction type in the history should match the one in the cache
+                        logger.LogWarning(
```
samuel-dev-dba: I don't really understand the logic I think we should prevent invalid data rather than trying to work around it.

keenesolutions: The factory create with Log an error if the earning type and transaction type are out of sync but the above code allows valid combinations and therefore refunds to be returned.

src/SFA.DAS.Payments.RequiredPayments.Application/Processors/PeriodisedRequiredPaymentEventFactory.cs
```diff
@@ -7,31 +8,70 @@
...
+
+        // ReSharper disable once IdentifierTypo
+        public PeriodisedRequiredPaymentEventFactory(IPaymentLogger logger)
+        {
+            this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
+        }
+
         public PeriodisedRequiredPaymentEvent Create(EarningType earningType, int transactionType)
         {
+            bool IsValidPaymentType<T>() where T : struct, IConvertible
```
samuel-dev-dba: Can we move this method out please

keenesolutions: I've change this to a private method.

keenesolutions: I've change this to a private method

ESFA-Pawan:
This has been tested using 0.0.5239

PR #840, opened 2019-12-09
Added indexes suggested by Azure in Prod.

MichaelYoung1981:


ESFA-Pawan:
Indexes created in DAS -  'DAS Payments V2-0.0.4953-R002'

PR #842, opened 2019-12-10
PV2 1683 Resolve unexpected clawback on restart after break

CarlHA:
Due to the complexity involved to get the earnings event matcher to function correctly with Planned breaks, following discussion with Tony we have decided to bypass the Event matcher for feature file PV2-216 as the earning events are covered by unit tests using an actual FM36 data file.

2 other feature files cover planned breaks - 296 is currently set to ignore and would require the feature file to be modified to be equivalent to 216 & 540 which has been set to ignore as the feature is incorrect and needs review by BA/Testers.


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Tests/English and Maths/PV2-540 - Levy learner English and Maths where planned duration is same as core program, planned break, other funding adjustment is 42.feature
```diff
@@ -47,7 +47,8 @@
  #| Funding Adjustment For Prior Learning |
  #| n/a                                   |
  #| 42% for remaining period|                                 | n/a                      |
-
+ @ignore
+ # Requires review by Test/BA to review as this refers to Next Academic year earnings which is incorrect.
```
keenesolutions: Is there a reason why this test is now being ignored?

src/SFA.DAS.Payments.AcceptanceTests.Core/Data/DataExtensions.cs
```diff
@@ -25,5 +26,16 @@ public static LearningAim ToLearningAim(this Course course)
...
+        public static T Clone<T>(this T source)
+        {
+            if (object.ReferenceEquals(source, null))
+            {
+                return default(T);
+            }
+
+            var deserialiseSettings = new JsonSerializerSettings{ObjectCreationHandling = ObjectCreationHandling.Replace};
+            return JsonConvert.DeserializeObject<T>(JsonConvert.SerializeObject(source), deserialiseSettings);
+        }
```
keenesolutions: Are we using the serializer to ensure we have a deep clone here?

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Tests/English and Maths/PV2-540 - Levy learner English and Maths where planned duration is same as core program, planned break, other funding adjustment is 42.feature
keenesolutions: Can it not be fixed if it's now broken due to new code?

CarlHA: As per the comment under the ignore flag.  This feature file was always invalid.

src/SFA.DAS.Payments.AcceptanceTests.Core/Data/DataExtensions.cs
CarlHA: Actually this code is no longer used and will be removed.

PR #844, opened 2019-12-16
Pv2 1710 Earnings Breakdown Tool

keenesolutions:


src/SFA.DAS.Payments.ProviderPayments.ServiceFabric/ApplicationPackageRoot/ApplicationManifest.xml
```diff
@@ -89,16 +89,16 @@
...
-    <!--<Service Name="SFA.DAS.Payments.ProviderPayments.LegacyExportService" ServicePackageActivationMode="ExclusiveProcess">
-      <StatefulService ServiceTypeName="SFA.DAS.Payments.ProviderPayments.LegacyExportServiceType" TargetReplicaSetSize="[LegacyExportService_TargetReplicaSetSize]" MinReplicaSetSize="[LegacyExportService_MinReplicaSetSize]">
-        <UniformInt64Partition PartitionCount="[LegacyExportService_PartitionCount]" LowKey="-9223372036854775808" HighKey="9223372036854775807" />
-      </StatefulService>
-    </Service>-->
     <!-- The section below creates instances of service types, when an instance of this 
          application type is created. You can also create one or more instances of service type using the 
          ServiceFabric PowerShell module.
          
          The attribute ServiceTypeName below must match the name defined in the imported ServiceManifest.xml file. -->
```
samuel-dev-dba: I don't think this change is required can you please revert it.

src/CalculationTools/EarningsComparer/Program.cs
```diff
@@ -0,0 +1,280 @@
...
+            return filterMode == FilterMode.Blacklist ?  
+                (joinedValues.Where(jv => !filterItems.Contains(jv.Ukprn)).ToList(), filterItems)
+                :
+                (joinedValues.Where(jv => filterItems.Contains(jv.Ukprn)).ToList(), filterItems);
+        }
+
+        private static List<long> GetFilterItems(FilterMode filterMode)
+        {
+            return filterMode == FilterMode.Blacklist
+                ? ResourceHelpers.LoadFilterValues(BlackListFile)
```
samuel-dev-dba: why is this filtering not done in SQL

keenesolutions: I'm using client filtering of records as I want to have the option of both white and black lists. We aren't talking about a huge amount of data here - worse case 3000 rows so client side filtering seemed like a good compromise since we'll most be be running this without any filtering at all.

keenesolutions:
This PR has been moved to the Payment-verification repo and is no longer required here.

PR #846, opened 2019-12-18
Fix Acceptance Test

samuel-dev-dba:


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndSteps.cs
```diff
@@ -428,6 +428,8 @@ public async Task WhenMonthEndIsTriggered()
         {
             await SendLevyMonthEnd().ConfigureAwait(false);
 
+            await Task.Delay(TimeSpan.FromSeconds(20));
```
MichaelYoung1981: Do we really want to delay for 20 seconds?

samuel-dev-dba: This is now removed

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/ApprovalsInterfaceSteps.cs
```diff
@@ -190,7 +190,7 @@ public void GivenTheApprenticeshipsHaveTheFollowingPriceEpisodes(Table table)
         }
 
         [When(@"the Approvals service notifies the Payments service of the apprenticeships")]
-        public void WhenTheApprovalsServiceNotifiesPaymentsVOfTheApprenticeships()
+        public async Task WhenTheApprovalsServiceNotifiesPaymentsVOfTheApprenticeshipsAsync()
```
abroodie: no need to append the method name with "Async"

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndSteps.cs
```diff
@@ -428,6 +429,8 @@ public async Task WhenMonthEndIsTriggered()
         {
             await SendLevyMonthEnd().ConfigureAwait(false);
 
+            await Task.Delay(TimeSpan.FromSeconds(20)); // wait for funding source service
```
abroodie: Why is this required?

samuel-dev-dba: We really need it for the test to simulate real-life workflow for month-end, until we can find a permanent probably using monitoring to determine when all funding source payments have been processed before triggering month-end.

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/ApprovalsInterfaceSteps.cs
samuel-dev-dba: Changed

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/EndToEndSteps.cs
abroodie: Agreed with Kola to remove this wait and also disable the code in the tests that is checking published payments as this isn't being used by any consumers yet.  Will need a ticket to address this properly.

PR #856, opened 2020-01-08
Pv2 1609 duplicate earning events

keenesolutions:


src/SFA.DAS.Payments.Application/Messaging/MessageLockTimeoutCheckBehaviour.cs
```diff
@@ -0,0 +1,40 @@
...
+
+
+        public override Task Invoke(ITransportReceiveContext context, Func<Task> next)
+        {
+            TimeSpan configTimeoutThreshold = configuration.MessageLockTimeoutThreshold;
+
+            var lockedUntilUtc = context.Extensions.Get<Message>().SystemProperties.LockedUntilUtc;
+            lockedUntilUtc -= configTimeoutThreshold;
+           
+            if (lockedUntilUtc >= DateTime.UtcNow)
```
samuel-dev-dba: we need a simple unit test for this and the logic should be <=

keenesolutions: I'm not sure why you think this should be <=?  If the locked until value is greater than noq then the lock hasn't expired and we should continue to process the message. If the locked time is less than now then message will now be available for other nodes to pick up and we shouldn't continue the execution on this node. 

I'm adjusting the lock timeout (to decrease the locktime) to give an additional buffer (contingency). This means that we are being pessimistic with out lock time and throwing the message out before it's expired.

keenesolutions: I've attempted to unit test this but ran into trouble due to the fact that Microsoft.Azure.ServiceBus.Message can only be created in an unread state and this can't be changed (without actual processing by the service bus pipline). The underlying property that changes the 'Read' state is internal and sealed. I spent an hour on this including atempting to set the locktimeout using reflection but decided that due to the simple nature of the logic that hacking in the unit test was worse.

PR #862, opened 2020-01-15
PV2-1747 Updated datamatch report to only use the most recent job id for each …

carlbm:
…provider


src/SFA.DAS.Payments.Database/Payments2/Views/Payments2.LatestSuccessfulJobs.sql
```diff
@@ -0,0 +1,18 @@
+﻿CREATE VIEW [Payments2].[LatestSuccessfulJobs]
+AS
+
+with validJobs (DcJobId, Ukprn)
+as
+(
+Select
+max(DcJobId),
```
keenesolutions: Can we be sure that the higher DCJobId is the latest one? Are these ever reset or recycled? I've seen instances of this before but this was probably in a test environment. Would ILR submission time be a safer to calculate this?

ESFA-Pawan:
This has been tested in SDW and looks ok.

PR #866, opened 2020-01-16
Pv2 1746 remove service that delete redundant payments

samuel-dev-dba:


src/SFA.DAS.Payments.Audit.EarningEventsService/Handlers/SubmissionSucceededHandler.cs
```diff
@@ -23,16 +23,18 @@ public SubmissionSucceededHandler(IPaymentLogger logger, IEarningEventSubmission
...
+            //    await processor.Process(message, CancellationToken.None).ConfigureAwait(false);
+            //    logger.LogInfo($"Finished processing submission succeeded event for earning events. message: {message.ToJson()}");
+            //}
+            //catch (Exception ex)
+            //{
+            //    logger.LogWarning($"Failed to remove previous submission earning events.  Error: {ex.Message}.  Ukprn: {message.Ukprn}, Collection period: {message.AcademicYear}-{message.CollectionPeriod}, Failed job: {message.JobId}.");
+            //    throw;
+            //}
+
+            await Task.CompletedTask;
```
carlbm: useless code...

src/SFA.DAS.Payments.Audit.FundingSourceService/Handlers/SubmissionSucceededHandler.cs
```diff
@@ -23,16 +23,18 @@ public SubmissionSucceededHandler(IPaymentLogger logger, IFundingSourceEventSubm
...
+            //    await processor.Process(message, CancellationToken.None).ConfigureAwait(false);
+            //    logger.LogInfo($"Finished processing submission succeeded event for funding source events. message: {message.ToJson()}");
+            //}
+            //catch (Exception ex)
+            //{
+            //    logger.LogWarning($"Failed to remove previous submission funding source events.  Error: {ex.Message}.  Ukprn: {message.Ukprn}, Collection period: {message.AcademicYear}-{message.CollectionPeriod}, Failed job: {message.JobId}.");
+            //    throw;
+            //}
+
+            await Task.CompletedTask;
```
carlbm: Can remove...

src/SFA.DAS.Payments.Audit.RequiredPaymentService/Handlers/SubmissionSucceededHandler.cs
```diff
@@ -22,16 +22,18 @@ public SubmissionSucceededHandler(IPaymentLogger logger, IRequiredPaymentEventSu
...
+            //    await processor.Process(message, CancellationToken.None).ConfigureAwait(false);
+            //    logger.LogInfo($"Finished processing submission succeeded event for required payment events. message: {message.ToJson()}");
+            //}
+            //catch (Exception ex)
+            //{
+            //    logger.LogWarning($"Failed to remove previous submission required payment events.  Error: {ex.Message}.  Ukprn: {message.Ukprn}, Collection period: {message.AcademicYear}-{message.CollectionPeriod}, Failed job: {message.JobId}.");
+            //    throw;
+            //}
+
+            await Task.CompletedTask;
```
carlbm: No need for this

carlbm:
I was under the impression that we should be removing the cleanup for both the success and failure events. After a conversation with Kola it is just the success events, with the failures continuing to be removed.

src/SFA.DAS.Payments.Audit.EarningEventsService/Handlers/SubmissionSucceededHandler.cs
samuel-dev-dba: I am sure there are a million ways to force things done your way than saying useless code.

samuel-dev-dba: The code is useful as you need to return something if you are within an async method

carlbm: No, you don't 😄 remove it and it will still compile and run

samuel-dev-dba: Then you will get a warning that not all code path returns a value. You really need to stop creating more warning in the system by not returning anything in an async method.

PR #874, opened 2020-01-28
PV2-1776 Prevent logging to SQL server

MichaelYoung1981:


src/SFA.DAS.Payments.AcceptanceTests.Components.Core/app.config
```diff
@@ -0,0 +1,55 @@
...
+        <assemblyIdentity name="System.ComponentModel.Annotations" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.2.1.0" newVersion="4.2.1.0" />
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="Autofac" publicKeyToken="17863af14b0044da" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.8.1.0" newVersion="4.8.1.0" />
+      </dependentAssembly>
+    </assemblyBinding>
+  </runtime>
+</configuration>
```
keenesolutions: Do we need these binding redirects or where they create automatically when upgrading some nuget packages?? 
If they weren't needed they can be ignored by selecting - options -> Nuget Package Manager -> Binding Redirects and checking "Skip applying binding redirects".

src/SFA.DAS.Payments.AcceptanceTests.Core/app.config
```diff
@@ -0,0 +1,55 @@
...
+        <assemblyIdentity name="Autofac" publicKeyToken="17863af14b0044da" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.8.1.0" newVersion="4.8.1.0" />
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="System.ValueTuple" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.0.3.0" newVersion="4.0.3.0" />
+      </dependentAssembly>
+    </assemblyBinding>
+  </runtime>
+</configuration>
```
keenesolutions: As above. Do we need these redirect?

PR #880, opened 2020-02-11
PV2-1764 Added a little more logging. Changed return value of earnings events …

carlbm:
…for invalid FM36 to return false to DC

### DIFF
--- a/src/SFA.DAS.Payments.EarningEvents.Application/Handlers/JobContextMessageHandler.cs
+++ b/src/SFA.DAS.Payments.EarningEvents.Application/Handlers/JobContextMessageHandler.cs
@@ -89,7 +89,8 @@ namespace SFA.DAS.Payments.EarningEvents.Application.Handlers
 
                     if (fm36Output == null)
                     {
-                        return true;
+                        logger.LogInfo("Could not find valid FM36 - aborting");
+                        return false;
                     }
 
                     cancellationToken.ThrowIfCancellationRequested();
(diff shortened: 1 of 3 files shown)


src/SFA.DAS.Payments.EarningEvents.Application/Handlers/JobContextMessageHandler.cs
```diff
@@ -89,7 +89,8 @@ public async Task<bool> HandleAsync(JobContextMessage message, CancellationToken
 
                     if (fm36Output == null)
                     {
-                        return true;
+                        logger.LogInfo("Could not find valid FM36 - aborting");
+                        return false;
```
keenesolutions: Is there a reason why we are changing logic here rather than just adding more logging??

src/SFA.DAS.Payments.EarningEvents.Application/Handlers/JobContextMessageHandler.cs
```diff
@@ -172,12 +173,14 @@ private async Task<bool> HandleSubmissionEvents(JobContextMessage message)
                 {
                     if (subscriptionMessage.Tasks.Any(t => t.Tasks.Contains(SubmissionJob.JobSuccess)))
                     {
+                        logger.LogInfo("Job contains a submission succeeded task - not processing any learners");
```
abroodie: The log is misleading.  These events don't really have anything to do with processing learners.  Also if you look inside the HandleSubmissionEvent code it already produces a log with more relevant information.

src/SFA.DAS.Payments.EarningEvents.Application/Handlers/JobContextMessageHandler.cs
```diff
@@ -172,12 +173,14 @@ private async Task<bool> HandleSubmissionEvents(JobContextMessage message)
...
                         await HandleSubmissionEvent<SubmissionSucceededEvent>(message);
                         return true;
                     }
 
                     if (subscriptionMessage.Tasks.Any(t => t.Tasks.Contains(SubmissionJob.JobFailure)))
                     {
+                        logger.LogInfo("Job contains a submission failed task - not processing any learners");
```
abroodie: The log is misleading.  These events don't really have anything to do with processing learners.  Also if you look inside the HandleSubmissionEvent code it already produces a log with more relevant information.

PR #889, opened 2020-02-13
Change Actor Ids to String

samuel-dev-dba:


src/SFA.DAS.Payments.RequiredPayments.RemovedLearnerService/RemovedLearnerService.cs
```diff
@@ -19,7 +19,11 @@ public class RemovedLearnerService : Actor, IRemovedLearnerService
         public RemovedLearnerService(ActorService actorService, ActorId actorId, IRemovedLearnerAimIdentificationService removedLearnerAimIdentificationService) : base(actorService, actorId)
         {
             this.removedLearnerAimIdentificationService = removedLearnerAimIdentificationService;
-            ukprn = actorId.GetLongId();
+
+            if (long.TryParse(actorId.GetStringId(), out ukprn))
```
carlbm: I think you're missing a `!`

samuel-dev-dba: Changed,

MichaelYoung1981:
> The missing `!` is a blocker, but there's also nothing in the implementation of the datalock actor to convert the string to an int. Either it's not required (in which case I would suggest removing all the superfluous logic in the other actors) or it is and this change won't work. Please investigate

I looked at that yesterday and the data lock service doesn't use the actor id, it picks the ULN off the message.

PR #902, opened 2020-02-21
Removed cascading deletes.

MichaelYoung1981:


MichaelYoung1981:
It won't fail, it readds it with the same name immediately afterwards.

PR #913, opened 2020-02-28
Pv2 1844 messages lock lost fix

hardik-desai-kcl:


src/SFA.DAS.Payments.Application/Messaging/DoNotProcessLockLostMessagesBehavior.cs
```diff
@@ -0,0 +1,43 @@
...
+                return next();
+            }
+            
+            var lockedUntilUtc = message.SystemProperties.LockedUntilUtc;
+            
+            if (lockedUntilUtc > DateTime.UtcNow) return next();
+            
+            logger.LogWarning($"Message Lock Expired: current time: {DateTime.UtcNow:G}, type: {message.UserProperties["NServiceBus.EnclosedMessageTypes"]}");
+
+            context.AbortReceiveOperation();
```
samuel-dev-dba: Why do we need to do this if we are going to renew the lock in  --> IIncomingLogicalMessageContext behavior.

carlbm: It's too late at this point. You can only renew a lock that hasn't timed out

PR #918, opened 2020-03-11
Pv2 1824 pv2 1826 data lock09 un happy path

hardik-desai-kcl:
test code refactor and added tests for 1826 and 1824


src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/FM36_ILR_Base_Steps.cs
```diff
@@ -36,12 +41,191 @@ public FM36_ILR_Base_Steps(FeatureContext context) : base(context)
...
 
+        [Given(@"there are 2 Commitments in DAS, Commitment A and Commitment B in collection period (.*)")]
+        public void GivenThereAreCommitmentsInDasCommitmentAAndCommitmentBInCollectionPeriodRCurrentAcademicYear(string collectionPeriodText)
+        {
+            GetFm36LearnerForCollectionPeriod(collectionPeriodText);
+        }
+
+        [Given(@"there is a single price episode in the ILR, (.*)")]
+        public void GivenThereIsASinglePriceEpisodeInTheIlrpe(string p0)
+        {
```
keenesolutions: Shouldn't we be checking this?

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/FM36_ILR_Base_Steps.cs
```diff
@@ -36,12 +41,191 @@ public FM36_ILR_Base_Steps(FeatureContext context) : base(context)
...
+
+            if (actualPriceEpisodeStartDate < apprenticeship.EstimatedStartDate) return;
+
+            apprenticeship.EstimatedStartDate = actualPriceEpisodeStartDate.AddDays(1);
+            apprenticeship.ApprenticeshipPriceEpisodes.Single().StartDate = actualPriceEpisodeStartDate.AddDays(1);
+            await testDataContext.SaveChangesAsync();
+        }
+
+        [Given(@"the start date in the (.*) is on or after the start date for Commitment (.*)")]
+        public async Task GivenTheStartDateInThePEIsOnOrAfterTheStartDateForCommitmentA(string priceEpisodeIdentifier, string commitmentIdentifier)
```
keenesolutions: A lot of code duplication here with the preceding method. The only difference is the date comparison. Can this be refactored to reduce duplication?

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/FM36_ILR_Base_Steps.cs
```diff
@@ -36,12 +41,191 @@ public FM36_ILR_Base_Steps(FeatureContext context) : base(context)
...
+            apprenticeship.EstimatedStartDate = actualPriceEpisodeStartDate.AddDays(-1);
+            apprenticeship.ApprenticeshipPriceEpisodes.Single().StartDate = actualPriceEpisodeStartDate.AddDays(-1);
+
+            await testDataContext.SaveChangesAsync();
+        }
+
+        [Given("end date of (.*) and the start date of (.*) occur in the same month")]
+        public void GivenEndDateOfEpisode1AndStartDateOfEpisode2OccurInTheSameMonth(string priceEpisodeIdentifier1, string priceEpisodeIdentifier2)
+        {
+            //NOTE: This is done else where in another step definition
```
keenesolutions: Should this be calling that code?

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/PV2_1825_Steps.cs
```diff
@@ -1,144 +1,32 @@
...
-
+            GetFm36LearnerForCollectionPeriod(collectionPeriodText);
         }
 
-        private bool HasDataLockErrorsForPriceEpisode(string priceEpisodeIdentifier, short academicYear)
+        [Given(@"the course in (.*) matches the course in Commitment (.*)")]
+        public void GivenTheCourseInPeMatchesTheCourseInCommitment(string priceEpisodeIdentifier, string commitmentIdentifier)
         {
-            return EarningEventsHelper.GetOnProgrammeDataLockErrorsForLearnerAndPriceEpisode(priceEpisodeIdentifier, academicYear, TestSession).Any();
+            //TODO: Copy course to apprenticeship
```
keenesolutions: Why is this not done?

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/PV2_1825_Steps.cs
```diff
@@ -1,144 +1,32 @@
...
-
-        [Then("there is a single match for (.*) with Commitment (.*)")]
-        public async Task ThereIsASingleMatchForPeWithCommitment(string priceEpisodeIdentifier, string commitmentIdentifier)
-        {
-            await WaitForIt(() => 
-                !HasDataLockErrorsForPriceEpisode(priceEpisodeIdentifier, short.Parse(TestSession.FM36Global.Year)) 
-                && HasSingleMatchForPriceEpisodeAndCommitment(priceEpisodeIdentifier),
-                "Failed to find a matching earning event and no datalocks."
-            );
+            //TODO Assert Course do not match
```
keenesolutions: Why is this not done?

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/FM36_ILR_Base_Steps.cs
hardik-desai-kcl: I did challenged this, but Tony said it was not necessary to validate input, in fact Mark you also said we can't validate input and gave me the sausage machine analogy, by the way these tests does not only have this one thing assumed to be correct there are many other input criteria that are assumed to be correct! example of the input being assumed correct is the start dates, if you mess around with the start dates these tests will fail!

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/FM36_ILR_Base_Steps.cs
hardik-desai-kcl: as described above

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/PV2_1825_Steps.cs
hardik-desai-kcl: as described above

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/PV2_1825_Steps.cs
hardik-desai-kcl: as described above

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/FM36_ILR_Base_Steps.cs
hardik-desai-kcl: done

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/FM36_ILR_Base_Steps.cs
keenesolutions: I appreciate what you are saying but the we **can** test for this and it won't change unless the inputs change (which will be static unless we change the feature code or the fm36). 

If we're not testing it I'm not sure why it's in the test. 

My sausage machine analogy was to say that other things can happen as a consequency of an action that we may not be concerned about in the testing (e.g. more than one data lock is raised). 

This is more that we are ensuring the inputs are correct, allowing us to test the expected outcome given those inputs. If we the expected outcome may change if there is more than one price episode then we should be testing that there is only one. If it has no bearing on the outcome then I'm not sure it should be in the test.

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/FM36_ILR_Base_Steps.cs
keenesolutions: We've lost the date comparsion in this refactor. We're hardcoding the date comparison from the first scenario.

PR #942, opened 2020-04-23
Pv2 1753 scheduled clean up of audit tables

hardik-desai-kcl:


src/SFA.DAS.Payments.ScheduledJobs.AcceptanceTests/Data/Entities/JobModel.cs
```diff
@@ -0,0 +1,66 @@
...
+        public DateTimeOffset? DcJobEndTime { get; set; }
+    }
+
+    public enum JobStatus : byte
+    {
+        InProgress = 1,
+        Completed,
+        CompletedWithErrors,
+        TimedOut,
+        DcTasksFailed
```
tom-gough, line 31: Enum with mixed explicit int value and auto numbering

src/SFA.DAS.Payments.ScheduledJobs.AcceptanceTests/Data/Entities/JobModel.cs
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public enum JobType : byte
+    {
+        EarningsJob = 1,
+        PeriodEndStartJob,
+        ComponentAcceptanceTestEarningsJob,
+        ComponentAcceptanceTestMonthEndJob,
+        PeriodEndRunJob,
+        PeriodEndStopJob
```
tom-gough, line 41: Enum with mixed explicit int value and auto numbering

src/SFA.DAS.Payments.ScheduledJobs.AcceptanceTests/Steps/AuditDataCleanUpSteps.cs
```diff
@@ -0,0 +1,422 @@
...
+        private readonly List<SubmissionData> submissions;
+        private readonly SubmissionDataContext submissionDataContext;
+        private readonly long ukprn;
+        private readonly Random randomNumberGenerator;
+
+        public AuditDataCleanUpSteps()
+        {
+            var config = new TestConfiguration();
+            submissionDataContext = new SubmissionDataContext(config.PaymentsConnectionString);
+            randomNumberGenerator = new Random(Guid.NewGuid().GetHashCode());
```
tom-gough, line 29: why used a seed value here?

src/SFA.DAS.Payments.ScheduledJobs/AuditDataCleanUp/AuditDataCleanUpService.cs
```diff
@@ -0,0 +1,133 @@
...
+            var jobsToBeDeleted = await dataContext.SubmissionJobsToBeDeleted.ToListAsync();
+            foreach (var job in jobsToBeDeleted)
+            {
+                try
+                {
+                    await dataContext.Database.ExecuteSqlCommandAsync(@"
+                    DELETE Payments2.EarningEventPeriod 
+                    FROM Payments2.EarningEventPeriod AS EEP 
+                        INNER JOIN Payments2.EarningEvent AS EE ON EEP.EarningEventId = EE.EventId 
+                    WHERE EE.JobId = @DcJobId",
```
tom-gough, line 41: Is it worth moving these strings to constants, either at the top of the class or in a separate static class?

src/SFA.DAS.Payments.ScheduledJobs/AuditDataCleanUp/DataLockAuditDataCleanUp.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        [FunctionName("DataLockEventAuditDataCleanUp")]
+        public static async Task DataLockEventAuditDataCleanUp([TimerTrigger("%AuditDataCleanUpSchedule%", RunOnStartup = false)] TimerInfo myTimer,
+                                                               [Inject] IAuditDataCleanUpService auditDataCleanUpService)
+        {
+            await auditDataCleanUpService.DataLockEventAuditDataCleanUp();
+        }
+
+        [FunctionName("HttpDataLockEventAuditDataCleanUp")]
+        public static async Task HttpDataLockEventAuditDataCleanUp([HttpTrigger(AuthorizationLevel.Function, "get", Route = null)] HttpRequest req,
```
tom-gough, line 21: Given this http call results in data being deleted/changed, should the verb be a put/post/delete rather than get?

src/SFA.DAS.Payments.ScheduledJobs/AuditDataCleanUp/AuditDataCleanUpService.cs
hardik-desai-kcl: because they are not being used anywhere else I think its best to keep it here and also keeping them here make it easier to read and understand code

src/SFA.DAS.Payments.ScheduledJobs/AuditDataCleanUp/DataLockAuditDataCleanUp.cs
hardik-desai-kcl: this is always going to run in scheduled mode so Http endpoint is for convenience only and 'Get' is easier to work with, if I change it to post or put or delete we have to use postman or something which is yet another pain point, Toby will appreciate it if we keep it simple

src/SFA.DAS.Payments.ScheduledJobs.AcceptanceTests/Data/Entities/JobModel.cs
hardik-desai-kcl: this is only for test but I have added numbering for all

src/SFA.DAS.Payments.ScheduledJobs.AcceptanceTests/Data/Entities/JobModel.cs
hardik-desai-kcl: this is only for test but I have added numbering for all

src/SFA.DAS.Payments.Database/Payments2/Views/Payments2.SubmissionJobsToBeDeleted.sql
```diff
@@ -0,0 +1,15 @@
...
+     UNION
+     SELECT J2.DCJobId
+     FROM Payments2.Job J1 --new
+     INNER JOIN Payments2.Job J2 --old
+     ON J1.Ukprn = J2.Ukprn --must have same Ukprn
+        AND J1.AcademicYear = J2.AcademicYear --must have same AcademicYear
+        AND J1.CollectionPeriod = J2.CollectionPeriod --must have same CollectionPeriod
+        AND J1.IlrSubmissionTime > J2.IlrSubmissionTime --new job must be in future
+     WHERE J2.JobId IS NOT NULL
+           AND J1.STATUS IN (2, 3, 4); -- get all old jobs where a new job exists with STATUS IN (2, 3, 4)
```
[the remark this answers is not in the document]
hardik-desai-kcl: done

src/SFA.DAS.Payments.ScheduledJobs.AcceptanceTests/Steps/AuditDataCleanUpSteps.cs
hardik-desai-kcl: https://stackoverflow.com/a/18267477

tom-gough: The default seed is based on system time, I think the reason they are suggesting new seeds from Guid.NewGuid() in that stack overflow post is because new instances of Random are being created in quick succession in a loop and therefore the default seed is the same each time, that isn't the case here.

https://docs.microsoft.com/en-us/dotnet/api/system.random.-ctor?view=netframework-4.8

Approved regardless as this is such a minor issue.

src/SFA.DAS.Payments.ScheduledJobs/AuditDataCleanUp/Extensions.cs
```diff
@@ -0,0 +1,25 @@
...
+using SFA.DAS.Payments.Model.Core.Audit;
+
+namespace SFA.DAS.Payments.ScheduledJobs.AuditDataCleanUp
+{
+    public static class Extensions
+    {
+        public static IEnumerable<SubmissionJobsToBeDeletedBatch> ToBatch(this IEnumerable<SubmissionJobsToBeDeletedModel> items, int maxItems)
+        {
+            return items.Select((submissionJobs, index) => new { submissionJobs, index })
+                .GroupBy(x => x.index / maxItems)
```
tom-gough, line 13: Not sure what the purpose of this .GroupBy is. Can the batching be achieved more simply with skip/take methods?

hardik-desai-kcl: because I am returning list if list grouping is the fastest way, skip take would require foreach here

src/SFA.DAS.Payments.ScheduledJobs/AuditDataCleanUp/AuditDataCleanUpService.cs
```diff
@@ -4,44 +4,61 @@
...
+        private readonly IEndpointInstanceFactory endpointInstanceFactory;
+        private readonly IScheduledJobsConfiguration config;
         private readonly IPaymentLogger paymentLogger;
 
-        public AuditDataCleanUpService([Inject] IPaymentsDataContext dataContext, [Inject] IPaymentLogger paymentLogger)
+        public AuditDataCleanUpService(
+            [Inject] IScheduledJobsConfiguration config,
+            [Inject] IPaymentsDataContext dataContext,
+            [Inject] IEndpointInstanceFactory endpointInstanceFactory,
+            [Inject] IPaymentLogger paymentLogger)
```
tom-gough, line 28: If you inject the service into the trigger, then you don't need these [Inject] markers as it will build the dependency tree when it instantiates the trigger and these will just work.

PR #961, opened 2020-06-11
Pv2 1844 b allow the processing of month end levy payments to last longer than the allocated time of 5 minutes

hardik-desai-kcl:


src/SFA.DAS.Payments.Application/Messaging/MessageTimeOutBehaviour.cs
```diff
@@ -0,0 +1,57 @@
...
+        public override async Task Invoke(ITransportReceiveContext context, Func<Task> next)
+        {
+            Message message;
+            try
+            {
+                message = context.Extensions.Get<Message>();
+            }
+            catch (KeyNotFoundException)
+            {
+                logger.LogWarning("Error getting Azure service bus Message from IIncomingLogicalMessageContext");
```
carlbm, line 34: The wording is slightly off. It's a `ITransportReceiveContext`  rather than a `IIncomingLogicalMessageContext`

src/SFA.DAS.Payments.Application/Messaging/MessageTimeOutBehaviour.cs
```diff
@@ -0,0 +1,57 @@
...
+                await next().ConfigureAwait(false);
+                return;
+            }
+
+            var lockedUntilUtc = message.SystemProperties.LockedUntilUtc;
+            if (lockedUntilUtc >= DateTime.UtcNow)
+            {
+                logger.LogWarning($"Message lock lost, discarding the message");
+                telemetry.TrackEvent("MessageLockLost", (lockedUntilUtc - DateTime.UtcNow).TotalSeconds);
+            }
```
carlbm, line 44: Don't we want to exit at this point?

src/SFA.DAS.Payments.Application/Messaging/MessageTimeOutBehaviour.cs
hardik-desai-kcl: fixed

src/SFA.DAS.Payments.Application/Messaging/MessageTimeOutBehaviour.cs
hardik-desai-kcl: if (lockedUntilUtc >= DateTime.UtcNow) and 
if (lockedUntilUtc <= DateTime.UtcNow.AddMinutes(1)) are inverse of each other so only one of those block will execute at a time so await next().ConfigureAwait(false); should be enough at line 54

PR #971, opened 2020-07-03
Add azure ad auth

msbaig:


hardik-desai-kcl:
another PR is created and associated with correct ticket

PR #975, opened 2020-07-10
PV2-2003 Added script to add the missing payments

carlbm:


tom-gough:
BEGIN TRAN with a commented out ROLLBACK and no commit seems bit messy for a checked in script?

PR #982, opened 2020-08-11
Pv2 2027 allow inline metrics to support multilple academic years

tom-gough:


src/SFA.DAS.Payments.Monitoring.AcceptanceTests/App.config
```diff
@@ -6,7 +6,7 @@
   <connectionStrings>
     <add name="StorageConnectionString" connectionString="UseDevelopmentStorage=true" />
     <add name="ServiceBusConnectionString" connectionString="" />
-    <add name="PaymentsConnectionString" connectionString="" />
+    <add name="PaymentsConnectionString" connectionString="Server=.;Database=SFA.DAS.Payments.Database;User Id=SFActor; Password=SFActor;" />
```
carlbm, line 9: Should have this in the repo

src/SFA.DAS.Payments.Monitoring.Metrics.Data/DcMetricsDataContextFactory.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public interface IDcMetricsDataContextFactory
+    {
+        IDcMetricsDataContext CreateContext(short academicYear);
+    }
+
+    public class DcMetricsDataContextFactory : IDcMetricsDataContextFactory
+    {
+        private readonly string connectionString;
+        private readonly Dictionary<short, IDcMetricsDataContext> contexts;
```
carlbm, line 14: It looks like you're calling this multiple times. Does it make sense to use a concurrent container?

PR #989, opened 2020-08-26
Pv2 2071 prevent period end run job executing twice

tom-gough:


src/SFA.DAS.Payments.AcceptanceTests.Core/Data/TestPaymentsDataContext.cs
```diff
@@ -105,12 +107,29 @@ public async Task ClearApprenticeshipData(long apprenticeshipId, long uln)
...
+                delete from Payments2.Job where JobId = {jobId}
+            ");
+        }
+
+        public bool JobExists(long jobId, short jobType)
+        {
+            using (var connection = (SqlConnection)Database.GetDbConnection())
+            {
+                connection.Open();
+                return new SqlCommand($"SELECT JobId FROM Payments2.Job WHERE DCJobId = {jobId} AND JobType = {jobType}", connection).ExecuteReader().HasRows;
```
hardik-desai-kcl, line 131: could you please check that only one record is returned from DB

hardik-desai-kcl (changes requested):

src/SFA.DAS.Payments.PeriodEnd.Application/Handlers/PeriodEndJobContextMessageHandler.cs
```diff
@@ -53,20 +58,43 @@ public async Task<bool> HandleAsync(JobContextMessage message, CancellationToken
...
+                {
+                    var endpointInstance = await endpointInstanceFactory.GetEndpointInstance();
+                    await endpointInstance.Publish(periodEndEvent);
+                    logger.LogInfo(
+                        $"Finished publishing the period end event. Name: {periodEndEvent.GetType().Name}, JobId: {periodEndEvent.JobId}, Collection Period: {periodEndEvent.CollectionPeriod.Period}-{periodEndEvent.CollectionPeriod.AcademicYear}.");
+                }
+                else
                 {
-                    await RecordPeriodEndJob(taskType, periodEndEvent).ConfigureAwait(false);
+                    var jobId = await jobsDataContext.GetJobId(GetJobType(taskType),
```
hardik-desai-kcl, line 74: rename jobId to existingJobId?

hardik-desai-kcl (changes requested):

src/SFA.DAS.Payments.AcceptanceTests.Core/Data/TestPaymentsDataContext.cs
tom-gough: I don't think that makes sense on this method, what would be do in the case where there is more than one job? Return false? That would be misleading.

Also there should never be a case where there is more than one job for the same DCJobId, if that is the case something fundamental has gone wrong, even before the fix in this ticket that was not happening.

hardik-desai-kcl (changes requested):

src/SFA.DAS.Payments.PeriodEnd.Application/Handlers/PeriodEndJobContextMessageHandler.cs
tom-gough: done

hardik-desai-kcl (approved):

PR #1023, opened 2020-11-26
 PV2-2215 only compare Apprenticeship details from last month

hardik-desai-kcl:


src/SFA.DAS.Payments.ScheduledJobs/Monitoring/ApprenticeshipData/ApprenticeshipDataService.cs
```diff
@@ -31,16 +32,22 @@ public ApprenticeshipDataService(IPaymentsDataContext paymentsDataContext, IComm
...
                 .GroupBy(g => g.PaymentStatus)
                 .Select(x => new { x.Key, Count = x.Count() })
                 .ToListAsync();
 
             var paymentsStatusTask = paymentsDataContext.Apprenticeship
                 .Where(a => a.Status == ApprenticeshipStatus.Active || a.Status == ApprenticeshipStatus.Stopped || a.Status == ApprenticeshipStatus.Paused)
+                .Where(a => a.CreationDate >= lastMonth)
                 .GroupBy(g => g.Status)
                 .Select(x => new { x.Key, Count = x.Count() })
                 .ToListAsync();
```
tom-gough, line 53: When we do the counts (just below in this class) is it possible that there might be some cases where the time discrepancy between EmployerAndProviderApprovedOn in Commitments and CreationDate in Payments causes a false difference in count? Would there be a benefit to comparing dates without time portions? You may still get a tiny number of edge cases submitted at midnight.

tom-gough (changes requested):

tom-gough (dismissed):

carlbm (dismissed):

carlbm (approved):

PR #1037, opened 2021-02-01
PV2-2332 Correct transfer payments that were not added in 1920

carlbm:


hardik-desai-kcl (approved):

hardik-desai-kcl (approved):

carlbm:
The most up-to-date script is now in Jira

PR #1042, opened 2021-03-19
Index added to payments table to improve performance of provider even…

tom-gough:
…ts query

### DIFF
--- a/src/SFA.DAS.Payments.Database/Payments2/Tables/Payment.sql
+++ b/src/SFA.DAS.Payments.Database/Payments2/Tables/Payment.sql
@@ -136,6 +136,6 @@ CREATE NONCLUSTERED INDEX [IX_Payment__CollectionPeriodCompletionPayments] on [P
 GO
 
 CREATE NONCLUSTERED INDEX [UX_Payment_ProviderEvents]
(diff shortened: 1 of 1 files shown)


src/SFA.DAS.Payments.Database/Payments2/Tables/Payment.sql
```diff
@@ -134,3 +134,8 @@ CREATE NONCLUSTERED INDEX [IX_Payment__CollectionPeriodCompletionPayments] on [P
 )
 
 GO
+
+CREATE NONCLUSTERED INDEX [UX_Payment_ProviderEvents]
+ON [Payments2].[Payment] ([CollectionPeriod],[AcademicYear],[FundingSource],[AccountId],[TransferSenderAccountId],[ApprenticeshipId])
```
carlbm, line 139: Considering that Funding Source should only be used for transfers I would move it to the end, or possibly before Apprenticeship Id, but definitely after AccountId 

Having looked through the queries I would change to:
CollectionPeriod, AcademicYear, AccountId, FundingSource, TransferSenderAccountId, ApprenticeshipId

There is also an option to use Ukprn in the query, but I'm not sure that it's used in the wild. There's no real way to integrate it with this index without removing support for transfers, which are definitely used

carlbm (changes requested):

carlbm (approved):

PR #1069, opened 2021-09-01
Pv2 2670 earnings adjustment statement

tom-gough:


src/SFA.DAS.Payments.ProviderAdjustments.Application/ProviderAdjustmentsProcessor.cs
```diff
@@ -40,25 +40,10 @@ public async Task ProcessEasAtMonthEnd(int academicYear, int collectionPeriod)
             var payments = calculator.CalculateDelta(historicPayments, currentPayments);
             logger.LogInfo($"Calculated {payments.Count} new payments");
 
-            PopulateCollectionPeriod(payments, academicYear, collectionPeriod);
+            calculator.PopulateCollectonPeriodForPayments(payments, academicYear, collectionPeriod);
```
hardik-desai-kcl, line 43: I would update the CalculateDelta and add academicYear, collectionPeriod as additional parameters and then pass them into CreatePayment methods and set the correct values as the object is getting created, that will avoid foreach

hardik-desai-kcl (changes requested):

hardik-desai-kcl (approved):

hardik-desai-kcl (approved):

PR #1084, opened 2021-11-23
PV2-2772-suppress-current-negative-dcEarnings-calculation

hardik-desai-kcl:


src/SFA.DAS.Payments.Monitoring.Metrics.Domain.UnitTests/PeriodEnd/Summary/PeriodEndProviderSummaryTests.cs
```diff
@@ -243,7 +243,7 @@ public void WhenGettingMetrics_AndThereAreNoNegativeEarnings_ThenNegativeEarning
             metrics.NegativeEarnings.ContractType2.Should().Be(0m);
         }
 
-        [TestCase(1000.00, 700.00)]
+        //[TestCase(1000.00, 700.00)]
```
tom-gough, line 246: Is it worth modifying this test such that we set the summary up with amounts for act1 & 2 negative earnings, then assert that the resulting earnings do not take them into account?

tom-gough (approved):

PR #1173, opened 2023-10-18
Pv2 3103 metrics generation failure slack alert

David-B-Read:


src/SFA.DAS.Payments.Monitoring.Metrics.Application/Submission/SubmissionMetricsService.cs
```diff
@@ -63,16 +66,22 @@ public async Task BuildMetrics(long ukprn, long jobId, short academicYear, byte
...
 
                 Task.WaitAny(dataTask, waitTask);
 
                 cancellationToken.ThrowIfCancellationRequested();
 
                 if (!dataTask.IsCompleted)
-                    throw new InvalidOperationException($"Took too long to get data for the submission metrics. Ukprn: {ukprn}, job: {jobId}, Collection period: {collectionPeriod}");
-                
+                {
+                    throw new SubmissionMetricsTimeoutException(
```
abroodie, line 81: Throwing a timeout exception is a bit misleading here as there could be any number of reasons why the metrics failed other than a sql timeout.  I think the ticket only talks about Metrics generation failure so probably best to stick to that domain terminology here. I would recommend throwing a SubmissionMetricsGenerationException.  Happy to discuss in the tech session though.

src/SFA.DAS.Payments.Monitoring.Metrics.Application.UnitTests/Submission/SubmissionMetricsGenerationTimeoutBehaviourTests.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        [Test]
+        public async Task Handles_Submission_Metrics_Timeout_Exception()
+        {
+            Func<Task> next = () => throw new SubmissionMetricsTimeoutException("Timed out");
+
+            await _sut.Invoke(_context, next);
+
+            var expectedMessage = $"Submission metrics generation timed out for UKPRN {_message.Ukprn}";
+            _telemetry.Verify(x => x.TrackEvent(expectedMessage,
```
abroodie, line 55: can these assertions be combined?

abroodie (changes requested):

src/SFA.DAS.Payments.Monitoring.Metrics.Application/Submission/SubmissionMetricsService.cs
David-B-Read: Renamed the exception and the custom behaviour classes to reference Generation instead of Timeout

src/SFA.DAS.Payments.Monitoring.Metrics.Application.UnitTests/Submission/SubmissionMetricsGenerationTimeoutBehaviourTests.cs
David-B-Read: Updated

abroodie (approved):

abroodie (approved):

LiamWright (approved):

PR #1174, opened 2023-11-07
PV2-3115 add extra metrics to Slack alert messages

David-B-Read:


owencampbell (approved):

abroodie (approved):

abroodie (approved):
The way SlackService consumes and uses the logger doesn't seem very elegant.

src/SFA.DAS.Payments.Monitoring.Alerts.Function/Helpers/ISlackAlertHelper.cs
```diff
@@ -7,11 +8,15 @@ public interface ISlackAlertHelper
     {
         public string GetEmoji(string severity);
 
-        public List<object> BuildSlackPayload(string alertEmoji,
+        public List<Block> BuildSlackPayload(string alertEmoji,
```
LiamWright, line 11: These are a lot of params. At some point, should this not be a class or condensed down in some way?

David-B-Read: Could be a candidate for a tech debt ticket, along with Tony's comment about the logger

LiamWright (approved):

PR #1231, opened 2025-12-18
Pv2 3783 update at remove apprenticeship updated handler

Winzar:
Adjusted existing acceptance tests to reflect records not being added to the reprocessing table upon update - as a part of the changes made by removing references to the ApprenticeshipUpdatedHandler in PeriodEnd (PV2-3782)

### DIFF
--- a/src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/ProvidersRequiringResubmissionSteps.cs
+++ b/src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/ProvidersRequiringResubmissionSteps.cs
@@ -55,15 +55,6 @@ namespace SFA.DAS.Payments.AcceptanceTests.EndToEnd.Steps
             }, options);
         }
 
-        [Then(@"new record will be added to the ProviderRequiringReprocessing table")]
-        public async Task ThenNewRecordWillBeAddedToTheProviderRequiringReprocessingTable()
-        {
-            await WaitForIt(() =>
-            {
-                return dataContext.ProvidersRequiringReprocessing.AnyAsync(x => x.Ukprn == TestSession.Ukprn);
-            }, $"Failed to find provider with matching ukprn: {TestSession.Ukprn} in ProviderRequiringReprocessing table ");
-        }
-
         [Then(@"there should not be any change to ProviderRequiringReprocessing table")]
         public async Task ThenThereShouldNotBeAnyChangeToProviderRequiringReprocessingTable()
         {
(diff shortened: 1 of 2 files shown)


LiamWright (dismissed):

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/ProvidersRequiringResubmissionSteps.cs
```diff
@@ -94,6 +85,7 @@ await MessageSession.Publish<SubmissionJobSucceeded>(m =>
             }, options);
         }
 
+        [Then("no record is added to the ProviderRequiringReprocessing table for that provider")]
         [Then(@"record for provider should be deleted from the ProviderRequiringReprocessing table")]
         public async Task ThenRecordForProviderShouldBeDeletedFromTheProviderRequiringReprocessingTable()
```
syed-wali, line 101: As we are using this "step" test/method to check if the ukprn exists, we should probably re-name it to sound generic rather than specific "deleted" check. Suggestion: CheckIfUkprnExistsInProviderRequiringReprocessingTable.

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/ProvidersRequiringResubmissionSteps.cs
syed-wali: This fail text is wrong. This will fail if a record was found, so the text should say that it found a record. 
Suggestion: "Failed - Unexpected provider with ukprn {TestSession.Ukprn} found in ProviderRequiringReprocessing table"

syed-wali (changes requested):

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/ProvidersRequiringResubmissionSteps.cs
Winzar: Additional note is that I need to add @ at the beginning of the string for consistency - don't believe it has an impact on the tests

Winzar: > As we are using this "step" test/method to check if the ukprn exists, we should probably re-name it to sound generic rather than specific "deleted" check. Suggestion: CheckIfUkprnExistsInProviderRequiringReprocessingTable.

Convention is that the method needs to start with Then, since it's a then method. Additionally, the convention within the file has other methods named by the text it corresponds to. I've currently set it to ThenNoRecordIsAddedToTheProviderRequiringReprocessingTableForThatProvider

Winzar: Spoke to Syed and resolved to have a separate method for the then step, which is to be named after the then step's instructions

src/SFA.DAS.Payments.AcceptanceTests.EndToEnd/Steps/ProvidersRequiringResubmissionSteps.cs
Winzar: Changes made

LiamWright (approved):

David-B-Read (approved):

syed-wali (approved):

